PR #9, opened 2017-02-09
TST: Add pandas BigQuery credentials to enable integration testing

### LINKED ISSUE #5: credential travis build
https://travis-ci.org/pydata/pandas-gbq already setup but needs to run full tests with credentials.

parthea:
Closes #5

The Dataset that is created during testing will include the current python version in the Dataset name in order to avoid conflicts when multiple python versions are tested simultaneously in the same BigQuery project.

@jreback


pandas_gbq/tests/test_gbq.py
```diff
@@ -16,19 +17,15 @@
...
 PROJECT_ID = None
 PRIVATE_KEY_JSON_PATH = None
 PRIVATE_KEY_JSON_CONTENTS = None
 
-if compat.PY3:
-    DATASET_ID = 'pydata_pandas_bq_testing_py3'
-else:
-    DATASET_ID = 'pydata_pandas_bq_testing_py2'
-
+DATASET_ID = 'pydata_pandas_gbq_py' + ''.join(map(str, sys.version_info))
```
jreback: sure, though we *may* want to have an env variable from the travis file (or default it to this if not on travis)

jreback:
lgtm. looks like 3.5/3.6 passed but others having issues.

jreback:
@parthea can you update?

parthea:
I'm on it

pandas_gbq/tests/test_gbq.py
```diff
@@ -64,6 +56,15 @@ def _in_travis_environment():
            'GBQ_PROJECT_ID' in os.environ
 
 
+def _get_dataset_id():
+    if _in_travis_environment():
+        python_version = os.environ.get('PYTHON').replace('.', '_')
```
jreback: maybe you should just generate an id here instead (say 10 random characters or something). this might collide if multiple travis workers are running (e.g. 2 PR's)

parthea: Done

jreback:
thanks @parthea

jreback:
@parthea actually master now failing :<

parthea:
Some gbq tests are flaky. The tests should pass if we retry the build. I'm going to create a new issue to capture flaky tests.

I'm also going to look at how we can handle the following error:
```
GenericGBQException: Reason: backendError, Message: Error encountered during execution. Retrying may solve the problem.
```

jreback:
@parthea no problem.

I think it might be wise to have a version of this decorator: https://github.com/pandas-dev/pandas/blob/master/pandas/util/testing.py#L2145

(and some associated machinery), though this should be somewhat simpler as it will be specific to the gbq service.

jreback:
https://pypi.python.org/pypi/flaky might be interesting

PR #14, opened 2017-02-26
Use name and type comparising when appending a dataframe into table

### LINKED ISSUE #13: to_gbq fails to append to table because of alleged schema mismatch
DF's schema returned by pandas.io.gbq.generate_bq_schema(df):
```
{'fields': [
{'type': 'TIMESTAMP', 'name': 'field1'}, 
{'type': 'STRING', 'name': 'field2'}, 
{'type': 'FLOAT', 'name': 'field3'}, 
{'type': 'STRING', 'name': 'field4'},
{'type': 'FLOAT', 'name': 'field5'},
{'type': 'STRING', 'name': 'field6'}, 
{'type': 'FLOAT', 'name': 'field7'}, 
{'type': 'STRING', 'name': 'field8'},
{'type': 'STRING', 'name': 'field9'}]}
```

Schema from BQ CLI:
```
|- field1: timestamp
|- field2: string
|- field3: float
|- field4: string
|- field5: float
|- field6: string
|- field7: float
|- field8: string
|- field9: string
```

`to_gbq(df,"dataset.table","project",if_exists="append",private_key=service_key_path)`
Throws:

```
  File "/lib/python2.7/site-packages/pandas/io/gbq.py", line 827, in to_gbq
    raise InvalidSchema("Please verify that the structure and "
pandas.io.gbq.InvalidSchema: Please verify that the structure and data types in the DataFrame match the schema of the destination table.
```

mremes:
I modified GbqConnector.verify_schema function to parse name and type from the remote schema (basically dropping mode) and include those in the compared fields.

Currently, when appending to a BQ table, comparison between the destination table's schema and a dataframe schema is done over superset of a BQ schema definition (name, type, mode) when _generate_bq_schema parses only name and type from a dataframe.

IMO it would be inconvenient to make the mode check in the module by generating completeness of columns (includes null values or not). So raising a generic GBQ error is more convenient here.

closes #13

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,8 +1,10 @@
 Changelog
 =========
 
-0.2.0 / 2017-?
+0.2.0 / 2017-02-26
 --------------
+
+Fixed an issue with appending to a BigQuery table where fields have modes (NULLABLE,REQUIRED,REPEATED). The changes concern solely the comparision of the local (DataFrame) and remote (BQ) schema in GbqConnector.verify_schema function. The fix is to omit other field attributes than name and type.
 
 0.1.2 / 2017-02-23
 ------------------
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 0.2.0 / 2017-02-26
 --------------
 
-Fixed an issue with appending to a BigQuery table where fields have modes (NULLABLE,REQUIRED,REPEATED). The changes concern solely the comparision of the local (DataFrame) and remote (BQ) schema in GbqConnector.verify_schema function. The fix is to omit other field attributes than name and type.
+Fixed an `issue <https://github.com/pydata/pandas-gbq/issues/13>` with appending to a BigQuery table where fields have modes (NULLABLE,REQUIRED,REPEATED). The changes concern solely the comparision of the local (DataFrame) and remote (BQ) schema in GbqConnector.verify_schema function. The fix is to omit other field attributes than name and type.
 
 0.1.2 / 2017-02-23
 ------------------
--- a/pandas_gbq/tests/test_gbq.py
+++ b/pandas_gbq/tests/test_gbq.py
@@ -1163,14 +1163,26 @@ class TestToGBQIntegrationWithServiceAccountKeyPath(tm.TestCase):
 
     def test_verify_schema_ignores_field_mode(self):
         test_id = "14"
-        test_schema_1 = {'fields': [{'name': 'A', 'type': 'FLOAT', 'mode': 'NULLABLE'},
-                                    {'name': 'B', 'type': 'FLOAT', 'mode': 'NULLABLE'},
-                                    {'name': 'C', 'type': 'STRING', 'mode': 'NULLABLE'},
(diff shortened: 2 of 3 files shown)


jreback:
can you add some tests?

jreback:
also can you add a release note to here: https://github.com/pydata/pandas-gbq/blob/master/docs/source/changelog.rst (I just added this)

mremes:
Added a test and supplied changelog.

jreback:
this is going to close #13 right?

mremes:
Yes, that's right.

docs/source/changelog.rst
```diff
@@ -1,9 +1,11 @@
 Changelog
 =========
 
-0.2.0 / 2017-?
+0.2.0 / 2017-02-26
 --------------
 
+Fixed an issue with appending to a BigQuery table where fields have modes (NULLABLE,REQUIRED,REPEATED). The changes concern solely the comparision of the local (DataFrame) and remote (BQ) schema in GbqConnector.verify_schema function. The fix is to omit other field attributes than name and type.
```
jreback: add ``(:issue:`13`)``.

jreback:
is this also the same soln as https://github.com/pandas-dev/pandas/issues/13086 (well is the issue the same)?

mremes:
Yes, it's the same issue. Though there was an issue with schema having descriptions, my case was just with modes – which violations could be made an exception case in some future time.

jreback:
@mremes is there anything you can take / tests from that issue for this to be more robust? (e.g. a test?)

pandas_gbq/tests/test_gbq.py
```diff
@@ -1161,6 +1161,34 @@ def test_upload_data_flexible_column_order(self):
...
+                                     'type': 'FLOAT'},
+                                    {'name': 'B',
+                                     'type': 'FLOAT'},
+                                    {'name': 'C',
+                                     'type': 'STRING'},
+                                    {'name': 'D',
+                                     'type': 'TIMESTAMP'}]}
+
+        self.table.create(TABLE_ID + test_id, test_schema_1)
+        self.assertTrue(self.sut.verify_schema(
```
jreback: just use 

``assert self.sut.verify_schema(.......), .....``

the ``self.assertTrue`` was a noseconvention, now using pytest so want to switch

mremes:
@jreback the fix should be a solid approach both to mine issue and the pandas-dev issue you referenced. Because the local schema is constructed from DF's column names and types, it's approariate to select only a name,field-subset of BQ fields when comparing.

I don't see a point in e.g. adding multiple discarded fields or anything like that.

pandas_gbq/tests/test_gbq.py
mremes: OK, I just looked the convention from the tests above..

jreback:
@mremes ok great. merging.

jreback:
thanks @mremes 

all set (though for some reason inter-sphinx links not working)....
https://pandas-gbq.readthedocs.io/en/latest/changelog.html#id2

pandas_gbq/tests/test_gbq.py
jreback: hah, I was going to change it on merge...but forgot...no worries

Issue #12, opened 2017-02-22
Printing rather than logging?


maxim-lian:
We're printing in addition to logging, when querying from BigQuery. This makes controlling the output much harder, aside from being un-idiomatic.

Printing in white, logging in red:

https://cloud.githubusercontent.com/assets/5635139/23176541/6028b884-f831-11e6-911a-48aa7741a4da.png

MaximilianR:
I think the ideal design is to remove the [`verbose` kwd](https://github.com/pandas-dev/pandas/blob/master/pandas/io/gbq.py#L146), and log with `INFO` or `WARNING` depending on the severity

jreback:
@MaximilianR that sounds good, as generally better to simply log, and people can connect to the console if desired. though I could see a kw controlling this too (as on an interactive query you simply want to have the printing and just easier), maybe ``verbose='print'``.

MaximilianR:
> as on an interactive query you simply want to have the printing and just easier

IPython shows logs assuming they are `WARNING` or above, by default, is that still necessary?

```python
In [1]: import logging

In [2]: logger = logging.getLogger()

In [3]: logger.info('test')

In [4]: logger.warning('test')
test
```

jreback:
oh maybe that's better. feel free to do a PR!

jreback:
@MaximilianR PR for this?

MaximilianR:
OK

jreback:
@MaximilianR bugging you!

MaximilianR:
Thank you for bugging me! I need to get better at pushing these over the line - I have a burst of enthusiasm and then finishing it can feel less exciting... I need a few days to get stuff back in order after traveling and then will aim to do this next week

tswast:
I've updated the changelog to include this change in https://github.com/pydata/pandas-gbq/pull/152

QuinRiva:
With the update and the changes to logging, I can't seem to get the progress status of downloading the results of the query.  Previously it would print an output after each page was downloaded (ever 30,000 rows or so) - no it only logs when everything has finished downloading.

I have tried using logging.INFO and logging.DEBUG, but neither gives me the desired output.  When downloading a table with millions of records, download progress is very useful.

`Requesting query... 
ok.
Query running...
Job ID: xxxx
Query running...
Got 242716 rows.

Total time taken 21.52 s.
Finished at 2018-05-23 13:51:13`

maxim-lian:
[...]

PR #18, opened 2017-03-01
ENH: log rather than print

maxim-lian:
closes https://github.com/pydata/pandas-gbq/issues/12


pandas_gbq/gbq.py
```diff
@@ -146,12 +148,17 @@ def __init__(self, project_id, reauth=False, verbose=False,
         _test_google_api_imports()
         self.project_id = project_id
         self.reauth = reauth
-        self.verbose = verbose
```
jreback: ok, can you change the default above to ``None``. I think will have to change this in pandas as well (its currently ``verbose=True``).

pandas_gbq/gbq.py
```diff
@@ -635,7 +632,7 @@ def _parse_entry(field_value, field_type):
 
 
 def read_gbq(query, project_id=None, index_col=None, col_order=None,
-             reauth=False, verbose=True, private_key=None, dialect='legacy',
+             reauth=False, verbose=None, private_key=None, dialect='legacy',
```
jreback: oh I c, you did change it at the top-level ok then.

jreback:
can you add an entry in ``doc/source/changelog.rst``

jreback:
cc @parthea ?

jreback:
@parthea can you have a look

parthea:
In https://github.com/pydata/pandas-gbq/blob/master/docs/source/reading.rst, the following note is no longer applicable:
```
.. note::

   You can toggle the verbose output via the ``verbose`` flag which defaults to ``True``.
```

In https://github.com/pydata/pandas-gbq/blob/master/docs/source/writing.rst, the following text is no longer applicable:

```
You can also see the progress of your post via the ``verbose`` flag which defaults to ``True``.
In [8]: to_gbq(df, 'my_dataset.my_table', projectid, chunksize=10000, verbose=True)

        Streaming Insert is 10% Complete
        Streaming Insert is 20% Complete
        Streaming Insert is 30% Complete
        Streaming Insert is 40% Complete
        Streaming Insert is 50% Complete
        Streaming Insert is 60% Complete
        Streaming Insert is 70% Complete
        Streaming Insert is 80% Complete
        Streaming Insert is 90% Complete
        Streaming Insert is 100% Complete
```

pandas_gbq/gbq.py
```diff
@@ -15,6 +15,8 @@
 from pandas.core.common import PandasError
 from pandas.compat import lzip, bytes_to_str
 
+logger = logging.getLogger(__name__)
```
parthea: I noticed the following message when running unit tests
```
No handlers could be found for logger "pandas_gbq.gbq"
```

Could we add a NullHandler to prevent the error when no logging configuration is provided? From the [logging docs](https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library),

> A do-nothing handler is included in the logging package: NullHandler (since Python 2.7). An instance of this handler could be added to the top-level logger of the logging namespace used by the library (if you want to prevent an error message being output to sys.stderr in the absence of logging configuration). 

For example,
```
logger.addHandler(logging.NullHandler())
```

On a separate note, could we enable info level logging by default when running tests (in test_gbq.py)?

pandas_gbq/gbq.py
```diff
@@ -409,10 +407,10 @@ def run_query(self, query, **kwargs):
 
         self._start_timer()
         try:
-            self._print('Requesting query... ', end="")
+            logger.info('Requesting query... ', end="")
```
parthea: Typo here that results in the following exception: `TypeError: _log() got an unexpected keyword argument 'end'`

```
    logger.info('Requesting query... ', end="")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <logging.Logger object at 0x7fc77b571150>, msg = 'Requesting query... ', args = ()
kwargs = {'end': ''}

    def info(self, msg, *args, **kwargs):
        """
            Log 'msg % args' with severity 'INFO'.
    
            To pass exception information, use the keyword argument exc_info with
            a true value, e.g.
    
            logger.info("Houston, we have a %s", "interesting problem", exc_info=1)
            """
        if self.isEnabledFor(INFO):
>           self._log(INFO, msg, args, **kwargs)
E           TypeError: _log() got an unexpected keyword argument 'end'
```

pandas_gbq/gbq.py
```diff
@@ -146,12 +148,17 @@ def __init__(self, project_id, reauth=False, verbose=False,
...
         self.reauth = reauth
-        self.verbose = verbose
         self.private_key = private_key
         self.dialect = dialect
         self.credentials = self.get_credentials()
         self.service = self.get_service()
 
+        if verbose is not None:
+                warnings.warn(
+                    "verbose is deprecated and will be removed in "
```
parthea: Could we add a unit test to exercise the deprecation warning? See `test_generate_bq_schema_deprecated` for similar example

jreback:
I also think would be nice to have a mini-section in the docs on how to setup logging properly.

pandas_gbq/gbq.py
MaximilianR: Generally the logging settings should be at the application level, rather than set by the code - so set at pytest. Though I don't think setting a `NullHandler` will interfere with anyone's settings

jreback: yep - ideally we test with it (to make sure it works)

and have some examples in docs in how to do it

jreback:
@MaximilianR if you wouldn't mind updating. I think after these can release 0.2.0

jreback:
@MaximilianR if you can update at some point :>

MaximilianR:
Yes thanks, I need to be better at pushing these over the finish line

jreback:
ping pong!

jreback:
ping

parthea:
@MaximilianR No rush on this one but it would be great to get this into the 0.2.0 release targeted for June 27.

MaximilianR:
Updated. I haven't added a logging guide or tests on the logging, but I'd hope this moves this forward regardless

pandas_gbq/gbq.py
```diff
@@ -4,14 +4,18 @@
...
+import logging
 
 import numpy as np
 
 from distutils.version import StrictVersion
 from pandas import compat, DataFrame, concat
 from pandas.compat import lzip, bytes_to_str
 
+logger = logging.getLogger(__name__)
+# silences warnings during tests
```
jreback: then this should be in the test module only

docs/source/changelog.rst
```diff
@@ -7,7 +7,9 @@ Changelog
 - Drop support for Python 3.4 (:issue:`40`)
 - The dataframe passed to ```.to_gbq(...., if_exists='append')``` needs to contain only a subset of the fields in the BigQuery schema. (:issue:`24`)
 - Use the `google-auth <https://google-auth.readthedocs.io/en/latest/>`__ library for authentication because oauth2client is deprecated. (:issue:`39`)
-- ``read_gbq`` now has a ``auth_local_webserver`` boolean argument for controlling whether to use web server or console flow when getting user credentials. Replaces `--noauth_local_webserver` command line argument (:issue:`35`)
+- ``read_gbq`` now has a ``auth_local_webserver`` boolean argument for controlling whether to use web server or console flow when getting user credentials.
+  Replaces `--noauth_local_webserver` command line argument (:issue:`35`)
+-
```
parthea: Could we also add an entry to the change log to say that the `verbose` parameter is deprecated and will be removed in a future version?

pandas_gbq/gbq.py
```diff
@@ -1186,13 +1195,14 @@ def delete(self, table_id):
 
 class _Dataset(GbqConnector):
 
-    def __init__(self, project_id, reauth=False, verbose=False,
+    def __init__(self, project_id, reauth=False, verbose=None,
```
parthea: Can we remove the verbose parameter here since _Dataset is internal?

pandas_gbq/gbq.py
```diff
@@ -1087,7 +1096,7 @@ def _generate_bq_schema(df, default_type='STRING'):
 
 class _Table(GbqConnector):
 
-    def __init__(self, project_id, dataset_id, reauth=False, verbose=False,
+    def __init__(self, project_id, dataset_id, reauth=False, verbose=None,
```
parthea: Can we remove the verbose parameter here since _Table is internal?

pandas_gbq/gbq.py
```diff
@@ -190,18 +194,23 @@ class TableCreationError(ValueError):
 class GbqConnector(object):
     scope = 'https://www.googleapis.com/auth/bigquery'
 
-    def __init__(self, project_id, reauth=False, verbose=False,
+    def __init__(self, project_id, reauth=False, verbose=None,
```
parthea: Can we remove the verbose parameter here since we don't support using `GbqConnector` directly (it's not documented)?

parthea:
Could you include an example in the docs to help users configure logging or point them to a reference?

Here is the code that I used to enable logging in my local testing:
```
import logging
import sys
logger = logging.getLogger('pandas_gbq.gbq')
logger.setLevel(logging.DEBUG)
logger.addHandler(logging.StreamHandler(stream=sys.stdout))
```

parthea:
@MaximilianR 
We lost the verbose logging on Travis with this change. Could you modify test_gbq.py so that we see logging on Travis? This may be a good place to enable logging: https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/tests/test_gbq.py#L93

jreback:
can you update for comments

pandas_gbq/gbq.py
```diff
@@ -807,7 +805,7 @@ def _parse_entry(field_value, field_type):
 
 
 def read_gbq(query, project_id=None, index_col=None, col_order=None,
-             reauth=False, verbose=True, private_key=None,
+             reauth=False, verbose=None, private_key=None,
```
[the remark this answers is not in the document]
maxim-lian: done

maxim-lian:
Updated and green

maxim-lian:
@parthea gentle ping

maxim-lian:
@parthea tell me if the pace of pings is annoying!

PR #24, opened 2017-04-06
When appending to a table, load if the dataframe contains a subset of the existing schema

mr-mcox:
## Purpose
Current behavior of `to_gbq` is fail if the schema of the new data is not equivalent to the current schema. However, this means that the load fails if the new data is missing columns that are present in the current schema. For instance, this may occur when the data source I am using to construct the dataframe only provides non-empty values. Rather than determining the current schema of the GBQ table and adding empty columns to my dataframe, I would like `to_gbq` to load my data if the columns in the dataframe are a subset of the current schema.

## Primary changes made

- Factoring a `schema` function out of `verify_schema` to support both `verify_schema` and  `schema_is_subset`
- `schema_is_subset` determines whether `local_schema` is a subset of `remote_schema`
- the append flag uses `schema_is_subset` rather than `verify_schema` to determine if the data can be loaded

## Auxiliary changes made

- `PROJECT_ID` etc are retrieved from an environment variable to facilitate local testing
- Running `test_gbq` through autopep8 added a row after two class names

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -12,6 +12,10 @@ Changelog
 
 - All gbq errors will simply be subclasses of ``ValueError`` and no longer inherit from the deprecated ``PandasError``.
 
+0.1.5 / 2017-04-20
+------------------
+- When using ```to_gbq``` if ```if_exists``` is set to ```append```, dataframe needs to contain only a subset of the fields in the BigQuery schema. GH#24
+
 0.1.4 / 2017-03-17
 ------------------
 
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,8 @@ Changelog
 
 - Resolve issue where the optional ``--noauth_local_webserver`` command line argument would not be propagated during the authentication process. (:issue:`35`)
 - Drop support for Python 3.4 (:issue:`40`)
+- When using ```to_gbq``` if ```if_exists``` is set to ```append```, dataframe needs to contain only a subset of the fields in the BigQuery schema. (:issue: `24`)
+
 
 0.1.6 / 2017-05-03
 ------------------
@@ -13,10 +15,6 @@ Changelog
 - All gbq errors will simply be subclasses of ``ValueError`` and no longer inherit from the deprecated ``PandasError``.
 
 0.1.5 / 2017-04-20
-------------------
-- When using ```to_gbq``` if ```if_exists``` is set to ```append```, dataframe needs to contain only a subset of the fields in the BigQuery schema. GH#24
-
-0.1.4 / 2017-03-17
 ------------------
 
 - ``InvalidIndexColumn`` will be raised instead of ``InvalidColumnOrder`` in ``read_gbq`` when the index column specified does not exist in the BigQuery schema. (:issue:`6`)
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,7 +6,7 @@ Changelog
 
 - Resolve issue where the optional ``--noauth_local_webserver`` command line argument would not be propagated during the authentication process. (:issue:`35`)
 - Drop support for Python 3.4 (:issue:`40`)
-- When using ```to_gbq``` if ```if_exists``` is set to ```append```, dataframe needs to contain only a subset of the fields in the BigQuery schema. (:issue: `24`)
+- The dataframe passed to ```.to_gbq(...., if_exists='append')``` needs to contain only a subset of the fields in the BigQuery schema. To support this, ```schema_is_subset``` tests whether a local dataframe is a subset of the BigQuery schema and ```schema``` returns the remote schema. (:issue:`24`)
 
 
 0.1.6 / 2017-05-03
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,7 +6,7 @@ Changelog
 
 - Resolve issue where the optional ``--noauth_local_webserver`` command line argument would not be propagated during the authentication process. (:issue:`35`)
 - Drop support for Python 3.4 (:issue:`40`)
-- The dataframe passed to ```.to_gbq(...., if_exists='append')``` needs to contain only a subset of the fields in the BigQuery schema. To support this, ```schema_is_subset``` tests whether a local dataframe is a subset of the BigQuery schema and ```schema``` returns the remote schema. (:issue:`24`)
+- The dataframe passed to ```.to_gbq(...., if_exists='append')``` needs to contain only a subset of the fields in the BigQuery schema. (:issue:`24`)
 
 
 0.1.6 / 2017-05-03
@@ -14,7 +14,7 @@ Changelog
 
 - All gbq errors will simply be subclasses of ``ValueError`` and no longer inherit from the deprecated ``PandasError``.
 
-0.1.5 / 2017-04-20
+0.1.4 / 2017-03-17
 ------------------
 
 - ``InvalidIndexColumn`` will be raised instead of ``InvalidColumnOrder`` in ``read_gbq`` when the index column specified does not exist in the BigQuery schema. (:issue:`6`)
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -557,7 +557,19 @@ class GbqConnector(object):
 
         self._print("\n")
 
-    def verify_schema(self, dataset_id, table_id, schema):
+    def schema(self, dataset_id, table_id):
+        """Retrieve the schema of the table
+
+        Obtain from BigQuery the field names and field types
+        for the table defined by the parameters
+
+        :param str dataset_id: Name of the BigQuery dataset for the table
+        :param str table_id: Name of the BigQuery table
+
+        :return: Fields representing the schema
+        :rtype: list of dicts
+        """
+
         try:
             from googleapiclient.errors import HttpError
         except:
@@ -573,14 +585,52 @@ class GbqConnector(object):
                               'type': field_remote['type']}
                              for field_remote in remote_schema['fields']]
 
-            fields_remote = set([json.dumps(field_remote)
-                                 for field_remote in remote_fields])
-            fields_local = set(json.dumps(field_local)
-                               for field_local in schema['fields'])
-
-            return fields_remote == fields_local
+            return remote_fields
         except HttpError as ex:
             self.process_http_error(ex)
+
+    def verify_schema(self, dataset_id, table_id, schema):
+        """Indicate whether schemas match exactly
+
+        Compare the BigQuery table identified in the parameters with
+        the schema passed in and indicate whether all fields in the former
+        are present in the latter. Order is not considered.
+
+        :param str dataset_id: Name of the BigQuery dataset for the table
+        :param str table_id: Name of the BigQuery table
+        :param list(dict) schema: Schema for comparison. Each item should have
+            a 'name' and a 'type'
+
+        :return: Whether the schemas match
+        :rtype: bool
+        """
+
+        fields_remote = sorted(self.schema(dataset_id, table_id),
+                               key=lambda x: x['name'])
+        fields_local = sorted(schema['fields'], key=lambda x: x['name'])
+
+        return fields_remote == fields_local
+
+    def schema_is_subset(self, dataset_id, table_id, schema):
+        """Indicate whether the schema to be uploaded is a subset
+
+        Compare the BigQuery table identified in the parameters with
+        the schema passed in and indicate whether a subset of the fields in
+        the former are present in the latter. Order is not considered.
+
+        :param str dataset_id: Name of the BigQuery dataset for the table
+        :param str table_id: Name of the BigQuery table
+        :param list(dict) schema: Schema for comparison. Each item should have
+            a 'name' and a 'type'
+
+        :return: Whether the passed schema is a subset
+        :rtype: bool
+        """
+
+        fields_remote = self.schema(dataset_id, table_id)
+        fields_local = schema['fields']
+
+        return all(field in fields_remote for field in fields_local)
 
     def delete_and_recreate_table(self, dataset_id, table_id, table_schema):
         delay = 0
@@ -844,7 +894,9 @@ def to_gbq(dataframe, destination_table, project_id, chunksize=10000,
             connector.delete_and_recreate_table(
                 dataset_id, table_id, table_schema)
         elif if_exists == 'append':
-            if not connector.verify_schema(dataset_id, table_id, table_schema):
+            if not connector.schema_is_subset(dataset_id,
+                                              table_id,
+                                              table_schema):
                 raise InvalidSchema("Please verify that the structure and "
                                     "data types in the DataFrame match the "
                                     "schema of the destination table.")
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -637,7 +637,7 @@ class GbqConnector(object):
             Name of the BigQuery dataset for the table
         table_id : str
             Name of the BigQuery table
-        schema : list(dict)  
+        schema : list(dict)
             Schema for comparison. Each item should have
             a 'name' and a 'type'
 
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -563,11 +563,17 @@ class GbqConnector(object):
         Obtain from BigQuery the field names and field types
         for the table defined by the parameters
 
-        :param str dataset_id: Name of the BigQuery dataset for the table
-        :param str table_id: Name of the BigQuery table
+        Parameters
+        ----------
+        dataset_id : str
+            Name of the BigQuery dataset for the table
+        table_id : str
+            Name of the BigQuery table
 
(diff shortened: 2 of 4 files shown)


pandas_gbq/tests/test_gbq.py
```diff
@@ -19,9 +19,9 @@
 import pandas.util.testing as tm
 from pandas.compat.numpy import np_datetime64_compat
 
-PROJECT_ID = None
-PRIVATE_KEY_JSON_PATH = None
-PRIVATE_KEY_JSON_CONTENTS = None
+PROJECT_ID = os.getenv('PROJECT_ID')
```
MaximilianR: I think the Google library pulls from `GOOGLE_CLOUD_PROJECT` (more confident about the newer library, which this package doesn't use yet). So you could either replicate that variable, or pass `None` through and the Google library should pick it up

pandas_gbq/tests/test_gbq.py
```diff
@@ -19,9 +19,9 @@
...
+PRIVATE_KEY_JSON_PATH = os.getenv('PRIVATE_KEY_JSON_PATH')
```
MaximilianR: Same as above for `GOOGLE_APPLICATION_CREDENTIALS`

pandas_gbq/tests/test_gbq.py
```diff
@@ -19,9 +19,9 @@
...
+PRIVATE_KEY_JSON_CONTENTS = os.getenv('PRIVATE_KEY_JSON_CONTENTS')
```
MaximilianR: (Contents doesn't have a Google version, so we do need to choose our own for that)

pandas_gbq/tests/test_gbq.py
```diff
@@ -1080,6 +1082,29 @@ def test_upload_data_if_table_exists_append(self):
...
+        # Initialize table with sample data
+        gbq.to_gbq(df, self.destination_table + test_id, _get_project_id(),
+                   chunksize=10000, private_key=_get_private_key_path())
+
+        # Test the if_exists parameter with value 'append'
+        gbq.to_gbq(df_subset_cols,
+                   self.destination_table + test_id, _get_project_id(),
+                   if_exists='append', private_key=_get_private_key_path())
+
+        sleep(30)  # <- Curses Google!!!
```
MaximilianR: We have a function `wait_for_job` which we copied from the Google docs; I'll put PR it in down the road (it relies on `google.cloud.bigquery`, so need to move some other code onto that)

pandas_gbq/tests/test_gbq.py
```diff
@@ -19,9 +19,9 @@
 import pandas.util.testing as tm
 from pandas.compat.numpy import np_datetime64_compat
 
-PROJECT_ID = None
-PRIVATE_KEY_JSON_PATH = None
-PRIVATE_KEY_JSON_CONTENTS = None
+PROJECT_ID = os.getenv('GOOGLE_CLOUD_PROJECT')
```
jreback: https://pandas-gbq.readthedocs.io/en/latest/contributing.html are the docs for how to test this

jreback:
is there any reason to change ``if_exists``? IOW, is there a reason *not* to append with a compat schema? (albeit one that is larger), ever?

pandas_gbq/gbq.py
```diff
@@ -558,7 +558,9 @@ def load_data(self, dataframe, dataset_id, table_id, chunksize):
 
         self._print("\n")
 
-    def verify_schema(self, dataset_id, table_id, schema):
+    def schema(self, dataset_id, table_id):
+        """Retrieve the schema of the table"""
```
jreback: can you add a doc-string

pandas_gbq/gbq.py
```diff
@@ -574,15 +576,26 @@ def verify_schema(self, dataset_id, table_id, schema):
...
-            fields_local = set(json.dumps(field_local)
-                               for field_local in schema['fields'])
-
-            return fields_remote == fields_local
+            return remote_fields
         except HttpError as ex:
             self.process_http_error(ex)
 
+    def verify_schema(self, dataset_id, table_id, schema):
+        fields_remote = set([json.dumps(field)
```
jreback: same here

pandas_gbq/gbq.py
```diff
@@ -574,15 +576,26 @@ def verify_schema(self, dataset_id, table_id, schema):
...
+                             for field in self.schema(dataset_id, table_id)])
+        fields_local = set(json.dumps(field_local)
+                           for field_local in schema['fields'])
+
+        return fields_remote == fields_local
+
+    def schema_is_subset(self, dataset_id, table_id, schema):
+        fields_remote = set([json.dumps(field)
```
jreback: and here

pandas_gbq/gbq.py
```diff
@@ -574,15 +576,26 @@ def verify_schema(self, dataset_id, table_id, schema):
...
+                             for field in self.schema(dataset_id, table_id)])
+        fields_local = set(json.dumps(field_local)
+                           for field_local in schema['fields'])
+
```
jreback: these routines should prob just take the result of ``self.schema(dataset_id, table_id)``, yes?

pandas_gbq/tests/test_gbq.py
```diff
@@ -1080,6 +1082,29 @@ def test_upload_data_if_table_exists_append(self):
                        _get_project_id(), if_exists='append',
                        private_key=_get_private_key_path())
 
+    def test_upload_subset_columns_if_table_exists_append(self):
+        test_id = "16"
```
jreback: can you add a comment with the issue ref number (this PR number is ok)

pandas_gbq/tests/test_gbq.py
```diff
@@ -1275,6 +1300,18 @@ def test_verify_schema_ignores_field_mode(self):
             self.dataset_prefix + "1", TABLE_ID + test_id, test_schema_2),
             'Expected schema to match')
 
+    def test_retrieve_schema(self):
+        test_id = "15"
```
jreback: comment of issue number

pandas_gbq/tests/test_gbq.py
```diff
@@ -1275,6 +1300,18 @@ def test_verify_schema_ignores_field_mode(self):
...
+        test_schema = {'fields': [{'name': 'A', 'type': 'FLOAT'},
+                                  {'name': 'B', 'type': 'FLOAT'},
+                                  {'name': 'C', 'type': 'STRING'},
+                                  {'name': 'D', 'type': 'TIMESTAMP'}]}
+
+        self.table.create(TABLE_ID + test_id, test_schema)
+        actual = self.sut.schema(self.dataset_prefix + "1", TABLE_ID + test_id)
+        expected = test_schema['fields']
+        assert expected == actual, 'Expected schema used to create table'
+
```
jreback: can you add some tests that exercise schema_is_subset

pandas_gbq/tests/test_gbq.py
parthea: I am ok with these environment variables (They are just for local testing). I agree with @jreback that is also necessary to follow the steps in the contributing docs to configure Travis-CI to run on your pandas-gbq fork. You only need to configure Travis-CI once. After that unit tests run automatically for your pandas-gbq fork after each push.

parthea:
> is there any reason to change if_exists? IOW, is there a reason not to append with a compat schema? (albeit one that is larger), ever?

@jreback I apologize if I missed the context for this comment. The default behaviour of the `if_exists` parameter is `fail` if a table already exists. Do we want to modify the default value of the `if_exists` parameter?

jreback:
@parthea 

>> is there any reason to change if_exists? IOW, is there a reason not to append with a compat schema? (albeit one that is larger), ever?

> @jreback I apologize if I missed the context for this comment. The default behaviour of the `if_exists` parameter is `fail` if a table already exists. Do we want to modify the default value of the `if_exists` parameter? 

what I mean is should this be a warning? you *could* have a 4th (maybe default state), like ``append_ok_schema_changes``. is that worthile?

parthea:
> what I mean is should this be a warning? you could have a 4th (maybe default state), like append_ok_schema_changes. is that worthile?

My initial thought is that we don't need to add another `if_exists` value. We could always add another value at a later stage if the community needs it.   While we are on the topic of the default state for the `if_exists` parameter, I do agree that the default value could be changed from `'fail'`  to `'append'`. I don't have a strong opinion about it.  @mr-mcox, thoughts ?

mr-mcox:
@parthea Regarding changing the `if_exists` default, I vote for keeping the default to `'fail'`. My rationale is that in BigQuery, removing records once you added them can be especially tricky. For the novice user, I'd want them to make a conscious choice to choose behavior that would add to an existing table.

@jreback has a good point in making the behavior of `'append'` also add new fields (not just adding a subset). This has been an occasional wish of mine and I may create a PR for that sometime in the future. However, I suspect that it may require some experimentation and require several more design considerations and ought to be out of the scope for this PR.

jreback:
ping, can you rebase and update to comments.

mr-mcox:
@jreback Sorry - I didn't realize that you were waiting on me for something. Can you be a little more specific about what you'd like me to do? Are you asking for me to take the five commits on this PR and rebase them to one, is there a comment that I neglected to address, or something else entirely?

jreback:
well needs a rebase, yes squashing is nice too (though not super necessary at this stage)

mr-mcox:
@jreback Alright - I've done a `git rebase upstream/master`, fixed a number of issues that resulted from that and squashed those commits. What would you like me to do next?

parthea:
@mr-mcox Thanks for the rebase. I'm really sorry for the delay in reviewing this pull request. I'm hoping to complete reviews for open pull requests this weekend.

pandas_gbq/gbq.py
```diff
@@ -557,7 +557,19 @@ def load_data(self, dataframe, dataset_id, table_id, chunksize):
...
         self._print("\n")
 
-    def verify_schema(self, dataset_id, table_id, schema):
+    def schema(self, dataset_id, table_id):
+        """Retrieve the schema of the table
+
+        Obtain from BigQuery the field names and field types
+        for the table defined by the parameters
+
+        :param str dataset_id: Name of the BigQuery dataset for the table
```
jreback: pls use the numpy-doc style (see all other doc-strings)

docs/source/changelog.rst
```diff
@@ -9,6 +9,10 @@ Changelog
 
 - All gbq errors will simply be subclasses of ``ValueError`` and no longer inherit from the deprecated ``PandasError``.
 
+0.1.5 / 2017-04-20
```
jreback: move to latest (0.1.7)

jreback: the issue reference is misformatted, see the other entries

pandas_gbq/gbq.py
```diff
@@ -573,15 +585,53 @@ def verify_schema(self, dataset_id, table_id, schema):
...
             self.process_http_error(ex)
 
+    def verify_schema(self, dataset_id, table_id, schema):
+        """Indicate whether schemas match exactly
+
+        Compare the BigQuery table identified in the parameters with
+        the schema passed in and indicate whether all fields in the former
+        are present in the latter. Order is not considered.
+
+        :param str dataset_id: Name of the BigQuery dataset for the table
```
jreback: same

pandas_gbq/gbq.py
```diff
@@ -573,15 +585,53 @@ def verify_schema(self, dataset_id, table_id, schema):
...
+        return fields_remote == fields_local
+
+    def schema_is_subset(self, dataset_id, table_id, schema):
+        """Indicate whether the schema to be uploaded is a subset
+
+        Compare the BigQuery table identified in the parameters with
+        the schema passed in and indicate whether a subset of the fields in
+        the former are present in the latter. Order is not considered.
+
+        :param str dataset_id: Name of the BigQuery dataset for the table
```
jreback: same

pandas_gbq/tests/test_gbq.py
```diff
@@ -1071,6 +1071,30 @@ def test_upload_data_if_table_exists_append(self):
                        _get_project_id(), if_exists='append',
                        private_key=_get_private_key_path())
 
+    def test_upload_subset_columns_if_table_exists_append(self):
+        # For pull request #24
```
jreback: be descriptive

pandas_gbq/tests/test_gbq.py
jreback: pls update as @MaximilianR suggests (look at other tests to see how this is done)

mr-mcox:
@jreback Sorry this took so long. I did another rebase, updated the doc strings, added some description to test cases and adjusted the change log per your requested changes. The only thing I didn't do was to change `test_gbq.py` around line 1100. If I'm following the conversation correctly, @MaximilianR 's suggestion was to use `wait_for_job` but first needed to PR that code in. I searched the repository for`wait_for_job` and didn't find this function. You mentioned looking at other tests - the test I modeled that one after `test_upload_data_if_table_exists_replace` still uses the `sleep(30)` workaround, so I'm wondering if I'm misunderstanding what you're looking for. Could you clarify? Is there anything else you need me to do at this time?

docs/source/changelog.rst
```diff
@@ -6,13 +6,15 @@ Changelog
 
 - Resolve issue where the optional ``--noauth_local_webserver`` command line argument would not be propagated during the authentication process. (:issue:`35`)
 - Drop support for Python 3.4 (:issue:`40`)
+- When using ```to_gbq``` if ```if_exists``` is set to ```append```, dataframe needs to contain only a subset of the fields in the BigQuery schema. (:issue: `24`)
```
jreback: no space after ``:issue:`24` ``, say more like:

``.to_gbq(...., if_exists='append')`` ...the rest of what you were saying

docs/source/changelog.rst
```diff
@@ -6,13 +6,15 @@ Changelog
...
+
 
 0.1.6 / 2017-05-03
 ------------------
 
 - All gbq errors will simply be subclasses of ``ValueError`` and no longer inherit from the deprecated ``PandasError``.
 
-0.1.4 / 2017-03-17
```
jreback: did you change this? there isn't a 1.5 release

pandas_gbq/gbq.py
```diff
@@ -557,7 +557,25 @@ def load_data(self, dataframe, dataset_id, table_id, chunksize):
 
         self._print("\n")
 
-    def verify_schema(self, dataset_id, table_id, schema):
+    def schema(self, dataset_id, table_id):
```
jreback: can you update the ``to_gbq`` doc-string as well (with expl of the new behavior of if_exists)

pandas_gbq/tests/test_gbq.py
jreback: @MaximilianR hmm, ``wait_for_job`` must be old.....

MaximilianR:
This is `wait_for_job`:

```

def wait_for_job(job):

    # from https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/cloud-client/snippets.py

    while True:
        if job.state == 'DONE':
            if job.error_result:
            return
        logger.info("Waiting for {} to complete".format(job))
        time.sleep(1)
```

jreback:
@MaximilianR thanks!

@mr-mcox if you want to add that as a utility function ok with that. pls run tests according to instructions and show the branch where they are passing.

(also ok to do this in another PR)

mr-mcox:
@jreback Good ideas on the updates to the changelog. You are correct that I removed 0.1.5 - this PR was the only change under that release and since there isn't 0.1.5 in the changelog on the master branch, I figured that it never happened and removed it entirely. I added `wait_for_job` with some minor tweaks, replaced the sleep functions across tests and the tests now complete in much less time. Since Travis CI on this page shows the tests on this branch passing, I'm not sure if you're looking for something else. That build is https://travis-ci.org/pydata/pandas-gbq/builds/240203212

mr-mcox:
Wait a sec - another build shows that I messed up calling `wait_for_job` - I'll fix that now.

mr-mcox:
@jreback I'm reverting that `wait_for_job` as it requires a job passed in. The job isn't returned by `to_gbq`, so I'm at a loss as to how to retrieve the job without refactoring `to_gbq`. @MaximilianR - what was your intention in using that utility? As I'm staring at `to_gbq`, I'm wondering if it is already waiting for the job to complete, though I vaguely recall from my early testing that the test failed without the wait.

MaximilianR:
Right, that's a good point:

- The current implementation of `to_gbq` uses streaming. This is OK, but slower (max 100k rows / sec), and more expensive. 
- The current implementation should complete very soon after returning - the data sits in a streaming buffer, and BQ queries that buffer before returning a result
- We use `upload_from_file` internally after writing to a CSV, which returns an async job, and so isn't ready immediately (though overall is faster)
- I would love to PR our stuff in, I've just been completely jammed here, and likely will be for another month
- Until we reconfigure `to_gbq` to use jobs, I would vote to allow `sleep`

jreback:
@parthea if you'd have a look

Issue #23, opened 2017-03-22
Structs lack proper names as dicts and arrays get turned into array of dicts


jasonqng:
Version 0.1.4

This query returns a improperly named dict:
```
q = """
select struct(a,b) col
from
(SELECT 1 a, 2 b)
"""
df = gbq.read_gbq(q, dialect='standard', verbose=False)
```
![image](https://cloud.githubusercontent.com/assets/3181961/24205707/8650b99e-0ef2-11e7-9954-fcbab8b2051d.png)

Compare with result from Big Query:
![image](https://cloud.githubusercontent.com/assets/3181961/24205739/9ddc6270-0ef2-11e7-863c-e3034ba3d34c.png)

An array of items also get turned into a arrays of dicts sometimes. For example:
```
q = """
select array_agg(a)
from
(select "1" a UNION ALL select "2" a)
"""
gbq.read_gbq(q, dialect='standard', verbose=False, project_id='project')
```
outputs:
![image](https://cloud.githubusercontent.com/assets/3181961/24205947/366b4ab0-0ef3-11e7-9969-6eb2011e9e4b.png)

Compare to Big Query:
![image](https://cloud.githubusercontent.com/assets/3181961/24205961/3dadd1f8-0ef3-11e7-8acf-482dddd13408.png)

These issues may or may not be related?

jreback:
@jasonqng if you wouldn't mind debugging this locally and see what the raw returns from bq are for these cases (and show here)?

IOW about here: https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/gbq.py#L607

jasonqng:
This is what I get back when running from `bq` command line:

```
creator-publishers> query --use_legacy_sql=False select struct(a,b) col from (SELECT 1 a, 2 b)
+-------------------+
|        col        |
+-------------------+
| {"a":"1","b":"2"} |
+-------------------+
```

jasonqng:
What I get for the second case, I get a proper list/array:
```
creator-publishers> query --use_legacy_sql=False select array_agg(a) from (select "1" a UNION ALL select "2" a)
+--------------+
|     f0_      |
+--------------+
| [u'1', u'2'] |
+--------------+
```

I'll tinker around a bit more tomorrow. But looks like this shouldn't be too hard to fix. 🤞

parthea:
Including the result from the api log as well (using `bq --apilog=-`):

```
{
 "kind": "bigquery#getQueryResultsResponse",
[...]

Issue #22, opened 2017-03-22
array_agg() fails because of TypeError


jasonqng:
Version: 0.1.4

When doing array_agg() on a column of INTs, you get a TypeError. The same query works fine in Big Query UI.

```
q = """
SELECT a, array_agg(mean_rank) ranks, count(*) ct
FROM `table`
group by a 
"""
gbq.read_gbq(q, dialect='standard', verbose=False, project_id='project')
```

Another example query to reproduce:
```
select array_agg(a)
from
(select 1 a)
```

Error message:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-b8c08e788608> in <module>()
      5 
      6 """
----> 7 gbq.read_gbq(q, dialect='standard', verbose=False, project_id='project')
      8

/home/jasonng/anaconda2/lib/python2.7/site-packages/pandas_gbq/gbq.pyc in read_gbq(query, project_id, index_col, col_order, reauth, verbose, private_key, dialect, **kwargs)
    725     while len(pages) > 0:
    726         page = pages.pop()
--> 727         dataframe_list.append(_parse_data(schema, page))
    728 
    729     if len(dataframe_list) > 0:

/home/jasonng/anaconda2/lib/python2.7/site-packages/pandas_gbq/gbq.pyc in _parse_data(schema, rows)
    621         for col_num, field_type in enumerate(col_types):
    622             field_value = _parse_entry(entries[col_num].get('v', ''),
--> 623                                        field_type)
    624             page_array[row_num][col_num] = field_value
    625 

/home/jasonng/anaconda2/lib/python2.7/site-packages/pandas_gbq/gbq.pyc in _parse_entry(field_value, field_type)
    631         return None
    632     if field_type == 'INTEGER':
--> 633         return int(field_value)
    634     elif field_type == 'FLOAT':
    635         return float(field_value)

TypeError: int() argument must be a string or a number, not 'list'
```

jasonqng:
Switching over to google-cloud-python would solve this. See comment and code here: https://github.com/pydata/pandas-gbq/issues/23#issuecomment-292672217

tswast:
Let's add a test case for this, and include it in the changelog.

jasonqng:
PR for test here: https://github.com/pydata/pandas-gbq/pull/101

PR #25, opened 2017-04-08
ENH: Convert read_gbq() function to use google-cloud-python

jasonqng:
# Description

I've rewritten the current `read_gbq()` function using `google-cloud-python`, which handles the naming of structs and arrays out of the box. For more discussion about this, see: https://github.com/pydata/pandas-gbq/issues/23. 

~However, because of the fact that google-cloud-python potentially uses different authentication flows and may break existing behavior, I've left the existing `read_gbq()` function and and named this new function `from_gbq()`. If in the future we are able to reconcile the authentication flows and/or decide to deprecate flows that are not supported in `google-cloud-python`, we can rename this to `read_gbq()`.~

UPDATE: As requested in comment by @jreback (https://github.com/pydata/pandas-gbq/pull/25/files/a763cf071813c836b7e00ae40ccf14e93e8fd72b#r110518161), I deleted old `read_gbq()` and named my new function `read_gbq()`, deleting all legacy functions and code.

Added in a few lines to requirements file, but I'll leave it to you @jreback to deal with conda dependency issues which you mentioned in Issue 23.

Let know if any questions or if any tests need to be written. You can confirm that it works by running the following:

```
q = """
select ROW_NUMBER() over () row_num, struct(a,b) col, c, d, c*d c_times_d, e
from
(select * from
    (SELECT 1 a, 2 b, null c, 0 d, 100 e)
    UNION ALL
    (SELECT 5 a, 6 b, 0 c, null d, 200 e)
    UNION ALL
    (SELECT 8 a, 9 b, 10.0 c, 10 d, 300 e)
)
"""
df = gbq.read_gbq(q, dialect='standard')
df
```

| row_num | col                | c    | d    | c_times_d | e   | 
|---------|--------------------|------|------|-----------|-----| 
| 2       | {u'a': 5, u'b': 6} | 0.0  | NaN  | NaN       | 200 | 
| 1       | {u'a': 1, u'b': 2} | NaN  | 0.0  | NaN       | 100 | 
| 3       | {u'a': 8, u'b': 9} | 10.0 | 10.0 | 100.0     | 300 | 

```
q = """
select array_agg(a) mylist
from
(select "1" a UNION ALL select "2" a)
"""
df = gbq.read_gbq(q, dialect='standard')
df
```

| mylist | 
|--------| 
| [1, 2] | 

```
q = """
select array_agg(struct(a,b)) col, f
from
(select * from
    (SELECT 1 a, 2 b, null c, 0 d, 100 e, "hello" f)
    UNION ALL
    (SELECT 5 a, 6 b, 0 c, null d, 200 e, "ok" f)
    UNION ALL
    (SELECT 8 a, 9 b, 10.0 c, 10 d, 300 e, "ok" f)
)
group by f
"""
df = gbq.read_gbq(q, dialect='standard')
df
```
| col                                      | f     | 
|------------------------------------------|-------| 
| [{u'a': 5, u'b': 6}, {u'a': 8, u'b': 9}] | ok    | 
| [{u'a': 1, u'b': 2}]                     | hello | 

Confirmed that `col_order` and `index_col` still work ~(feel free to pull that out into a separate function since there's now redundant code with `read_gbq()`)~, and I removed the type conversion lines which appear to be unnecessary (google-cloud-python and/or pandas appears to do the necessary type conversion automatically, even if there are nulls; can confirm by examining the datatypes in the resulting dataframes).


pandas_gbq/gbq.py
```diff
@@ -767,6 +770,116 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
 
     return final_df
 
+def from_gbq(query, project_id=None, index_col=None, col_order=None,
```
jreback: this should simply replace ``read_gbq``, changing the top-level API is a non-starter.

jasonqng: Sure, I can rename mine to `read_gbq` and add in the query length/bytes processed info. Want me to delete the old `read_gbq` and related code or just rename it?

jreback:
this need to pass all of the original test
it's simply an implementation change

pandas_gbq/gbq.py
```diff
@@ -603,106 +458,56 @@ def delete_and_recreate_table(self, dataset_id, table_id, table_schema):
...
+    The easiest is to download a service account JSON keyfile and point to it using 
+    an environment variable:
+    `$ export GOOGLE_APPLICATION_CREDENTIALS="/path/to/keyfile.json"`
 
     Parameters
     ----------
     query : str
         SQL-Like Query to return data values
-    project_id : str
+    project_id : str (optional)
```
jreback: how is this optional?

pandas_gbq/gbq.py
```diff
@@ -603,106 +458,56 @@ def delete_and_recreate_table(self, dataset_id, table_id, table_schema):
...
         Google BigQuery Account project ID.
     index_col : str (optional)
         Name of result column to use for index in results DataFrame
     col_order : list(str) (optional)
         List of BigQuery column names in the desired order for results
         DataFrame
-    reauth : boolean (default False)
```
jreback: is this now not needed? if so, then simply mark it as deprecated (and raise a warning if its passed)

pandas_gbq/gbq.py
```diff
@@ -603,106 +458,56 @@ def delete_and_recreate_table(self, dataset_id, table_id, table_schema):
...
+        Path to service account private key in JSON format. If none is provided,
+        will default to the GOOGLE_APPLICATION_CREDENTIALS environment variable
+        or another form of authentication (see above)
     dialect : {'legacy', 'standard'}, default 'legacy'
         'legacy' : Use BigQuery's legacy SQL dialect.
         'standard' : Use BigQuery's standard SQL (beta), which is
         compliant with the SQL 2011 standard. For more information
         see `BigQuery SQL Reference
         <https://cloud.google.com/bigquery/sql-reference/>`__
-
```
jreback: why the change here?

jreback:
@jasonqng we have to be careful about back-compat here.

jreback:
you need to update the ``ci/requirements-*.pip`` files, removing the old requirements and adding the new

pandas_gbq/gbq.py
jasonqng: >project (str) – the project which the client acts on behalf of. Will be passed when creating a dataset / job. If not passed, falls back to the default inferred from the environment.

https://googlecloudplatform.github.io/google-cloud-python/stable/bigquery-client.html#google.cloud.bigquery.client.Client

pandas_gbq/gbq.py
jasonqng: From what I can gather from the discussion here (https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2765) passing an arbitrary JSON of configuration settings isn't supported in the way it was in the previous python api. As such, we might as well make the passing of configuration settings a little easier with a dict like so, but happy to consider alternatives.

pandas_gbq/gbq.py
jasonqng: Yeah, not sure how to implement or if such behavior can be replicated in new api (or is even desired). If any folks have thought, let know. Otherwise, if I can't, will do as you suggest and raise a warning if reauth is passed.

jasonqng:
@jreback Yeah, the back compatibility issues with the authentication is partly why I suggested writing it as a new function, but hopefully we can replicate a form of the pop-up authentication->refresh token with the new api (https://googlecloudplatform.github.io/google-cloud-python/stable/google-cloud-auth.html#user-accounts-3-legged-oauth-2-0-with-a-refresh-token). I might need some help with that if others are more familiar with it. Almost everything else should carry over, so I'm not too concerned with compatibility otherwise.

pandas_gbq/gbq.py
```diff
@@ -711,28 +516,79 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+        for unit in ['', 'K', 'M', 'G', 'T', 'P', 'E', 'Z']:
+            if abs(num) < 1024.0:
+                return fmt % (num, unit, suffix)
+            num /= 1024.0
+        return fmt % (num, 'Y', suffix)
 
-    if dialect not in ('legacy', 'standard'):
-        raise ValueError("'{0}' is not valid for dialect".format(dialect))
+    if private_key:
+        os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = private_key
```
thejens: Not sure if it is nice to set environ vars, can't you just include a custom reference to the private_key, if it exists, when creating the Client on row 540?

requirements.txt
```diff
@@ -2,3 +2,5 @@ pandas
 httplib2
 google-api-python-client
 oauth2client
+google-cloud
```
thejens: google-cloud is not the most stable package (I've noticed), I'd require a specific version

pandas_gbq/gbq.py
jasonqng: Good suggestion, will do.

pandas_gbq/gbq.py
```diff
@@ -603,106 +458,56 @@ def delete_and_recreate_table(self, dataset_id, table_id, table_schema):
         table.create(table_id, table_schema)
         sleep(delay)
 
-
-def _parse_data(schema, rows):
```
parthea: In test_gbq.py, there are several tests that have `gbq._parse_data(...)`. Could you update `test_gbq.py` as well? See the tests with prefix `test_should_return_bigquery*(...)`.

requirements.txt
jreback: that's fine, you can pip specifically


```google-cloud=0.24.0`` (or whatever)

File: pandas_gbq/gbq.py (hunk 7 above)
Comment (jasonqng) [inline pandas_gbq/gbq.py]: Yup, will tackle this weekend hopefully. Thanks!

File: requirements.txt (hunk 6 above)
Comment (MaximilianR) [inline requirements.txt]: You can also be more specific and only require the bigquery library with `google-cloud-bigquery`: https://github.com/GoogleCloudPlatform/google-cloud-python/tree/master/bigquery

File: pandas_gbq/gbq.py (hunk 4 above)
Comment (tswast) [inline pandas_gbq/gbq.py]: FYI: this is one reason I'd prefer to use google-auth library directly. https://github.com/pydata/pandas-gbq/issues/26

Comment (MaximilianR) [inline pandas_gbq/gbq.py]: @tswast @jasonqng 
You can get a `Credentials` object from the json with `google.oauth2.service_account.Credentials.from_service_account_info(json.loads(key))`

Forgive me for not PRing this sort of thing in directly - I'm totally jammed at the moment (forgive me @jreback too...). But happy to help with any questions on this stuff - we've run through a lot of it over here

File: requirements.txt (hunk 6 above)
Comment (tswast) [inline requirements.txt]: +1 to using `google-cloud-bigquery`.

Also, the `google-cloud-bigquery` package is still beta, meaning there will likely be breaking changes, so we'd have to pin a very specific version number to maintain stability.

**Comment by tswast** [comment]
I recommend closing this PR.

The `google-cloud-bigquery` library is not yet 1.0. Breaking changes are very likely as we (Google) get the libraries in good shape for a 1.0 release (no specific timeline on that yet). There are also a few things yet-to-be-implemented in `google-cloud-bigquery` such as retry logic, which are in the "API client library".

We should revisit this after the `google-cloud-bigquery` library goes 1.0.

**Comment by jasonqng** [comment]
@jreback In light of @tswast's comment, should we just close this or should we go back to building this as a separate function (e.g. `from_gbq()`) as I had it before? We could clearly mark it as experimental. For my own selfish sake, I'd vote for latter just so I can start using it in the mean time before google-cloud API goes 1.0, but obviously defer to others here.

**Comment by jreback** [comment]
we can easily just pin to a specific version of api stability is a concern but in general i don't see this as a big deal

no reason to wait for a 1.0

Pandas itself is not even 1.0 and lots of people use / depend on it

File: pandas_gbq/gbq.py [hunk 8]
```diff
@@ -603,106 +458,56 @@ def delete_and_recreate_table(self, dataset_id, table_id, table_schema):
...
-
-    Authentication to the Google BigQuery service is via OAuth 2.0.
-
-    - If "private_key" is not provided:
+    <https://googlecloudplatform.github.io/google-cloud-python/stable/>`__
 
-      By default "application default credentials" are used.
-
-      If default application credentials are not found or are restrictive,
-      user account credentials are used. In this case, you will be asked to
```
tswast: Are you sure you want to remove the user account credentials method? I believe it can be made to work with the google-cloud-biguery library. I'm working on adding a sample to the Google docs that does the user-auth flow.

jasonqng: That would be awesome. Do feel free to point to the doc when it's added. Sorry, been a bit distracted with work but hope to address all comments next week (and agree with @jreback's [comment](https://github.com/pydata/pandas-gbq/pull/25#issuecomment-296320690) to move forward with this). Thanks.

tswast: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/925

That pull request adds a sample which uses https://pypi.python.org/pypi/google-auth-oauthlib to run a query with user account credentials.

I'll be writing some tests and docs around it, but the code should be easy enough to follow.

tswast:
@jasonqng You'll probably want to rebase this change after https://github.com/pydata/pandas-gbq/pull/39 gets merged. The `google-auth` library is used by both `google-api-python-client` and `google-cloud-python`, so you should be able to reuse my changes for user and service account authentication.

jreback:
is this PR relevant after #39 ?

@tswast

parthea:
We still have #23 as an open issue. It would be great to move this forward to address that. I recently added a conda recipe for `google-cloud-bigquery`.
https://github.com/conda-forge/google-cloud-bigquery-feedstock

jreback:
@parthea ok is it worth waiting on this for 0.2.0 to avoid even more changes?

parthea:
My initial thought is that the milestone for this PR should be 0.3.0 as thorough testing is required. I think ultimately it depends on how soon we can begin testing this PR and whether we are in a hurry to release 0.2.0.

@jasonqng Please could you rebase ?

tswast:
Yeah, this PR is still relevant. #39 moves pandas-gbq to use `google-auth` but still use the Google API client libraries. This PR is to move it to the Google Cloud client libraries (which incidentally also uses `google-auth`, so rebasing would be really helpful in completing the work required to get this working properly)

jasonqng:
@tswast @parthea @jreback Sorry, been swamped these past few months. Hope to scratch out some time this week to incorporate all comments (and also get it to working with queries with large results, which it currently fails with). Just checking, any particular reason to rebase vs a merge? Happy to do former, just haven't done rebase on any collaborative projects so this would be first time. (Haha, worst case scenario, I mess up my branch and I just rewrite and open a new PR branched off new master.)

parthea:
@jasonqng Thanks for taking care of this! Rebase is preferred because it will allow you to add commits on top of the latest master which is much nicer to look at during code review.

GregCT:
Hi,
Just wondering if @jasonqng has already made any progress rebasing this?
If not I'd be keen to help finish it off and get the changes reviewed and merged.

jasonqng:
@GregCT Thanks for offering! Been meaning to wrap this up, but been swamped. I'd love some help finishing this. I can add you as a contributor so you can push directly. Look forward to working together on this if you're able!

jasonqng:
@tswast @jreback @parthea Have successfully rebased and confirmed that the function still works (just successfully queried for 100,000 rows and got all rows in dataframe). Will work on updating the authentication hopefully this week. Look forward to pushing this toward the finish line, though I might need help with tests. Thanks for your patience and do let me know if any other issues I should be aware of in order to complete this.

requirements.txt
```diff
@@ -4,3 +4,5 @@ google-api-python-client
 google-auth
 google-auth-httplib2
 google-auth-oauthlib
+google-cloud
```
MaximilianR: while this is fine, I _think_ this could be `google-cloud-bigquery`, which is smaller?

tswast: :+1:  to using `google-cloud-bigquery` instead of `google-cloud`

pandas_gbq/gbq.py
```diff
@@ -1014,6 +993,147 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
 
     return final_df
 
+def read_gbq(query, project_id=None, index_col=None, col_order=None, verbose=True,
```
tswast: Looks like there was maybe a problem with the rebase. This is a duplicate definition of `read_gbq`.

jasonqng: Wow, totally didn't notice since calling `read_gbq` uses the new one I wrote, which is further down. Kinda surprised python didn't complain, but maybe I shouldn't. Removed old one in latest commit, thanks for catching!

requirements.txt
jasonqng: Verified on a clean environment that `google-cloud-bigquery==0.25.0` is enough. However, `0.26.0` throws the following error:

```
  File "gbq.py", line 960, in read_gbq
    query_results = query_job.results()
AttributeError: 'QueryJob' object has no attribute 'results'
```

Pinning to `0.25.0` for now, thanks, great suggestion.

pandas_gbq/gbq.py
jasonqng: @thejens Fixed and verified it works with my service account json key.

requirements.txt
jasonqng: Fixed and pinned to google-cloud-bigquery to 0.25.0 since 0.26.0 breaks things

requirements.txt
MaximilianR: Thanks for checking @jasonqng 

Although that's a bit concerning that this breaks for recent versions of that library - that's going to force people to roll back upgrades. Others can come to a verdict on whether that's a showstopper, but at least worth seeing if we can make it compatible with a newer version?

pandas_gbq/gbq.py
jasonqng: @MaximilianR solved the auth with a key in latest commit. is this other concern about the config settings still an open issue to be resolved in this PR?

pandas_gbq/gbq.py
```diff
@@ -872,82 +851,58 @@ def _parse_entry(field_value, field_type):
...
-    reauth : boolean (default False)
-        Force Google BigQuery to reauthenticate the user. This is useful
-        if multiple accounts are used.
     verbose : boolean (default True)
         Verbose output
     private_key : str (optional)
-        Service account private key in JSON format. Can be file path
-        or string contents. This is useful for remote server
-        authentication (eg. jupyter iPython notebook on remote host)
-    auth_local_webserver : boolean, default False
```
tswast: You should be able to keep this. The existing auth code on master should work seamlessly with both the Google API client libs and the Google Cloud client libs.

There's an example of using the local web server for auth here: https://cloud.google.com/bigquery/docs/authentication/end-user-installed

MaximilianR:
I think this https://github.com/pydata/pandas-gbq/pull/25#discussion_r142777163 may be hidden as it's on a changed piece of code - repeating here: 

> Thanks for checking @jasonqng 

>Although that's a bit concerning that this breaks for recent versions of that library - that's going to force people to roll back upgrades. Others can come to a verdict on whether that's a showstopper, but at least worth seeing if we can make it compatible with a newer version?

MaximilianR:
> I can't for the life of me figure out why pytest won't run

Can you post pip list?

pandas_gbq/gbq.py
MaximilianR: I think this is solved now, thank you @jasonqng

tswast: You'll also need to pin the version in `setup.py`.

requirements.txt
tswast: You'll also need to pin the version in `setup.py`.

jasonqng:
@MaximilianR this is on a clean environment; only change was I intentionally downgraded python-dateutil to 2.6.0 from 2.6.1 in case that helped, but it didn't:

```
cachetools (2.0.1)
certifi (2017.7.27.1)
chardet (3.0.4)
google-api-python-client (1.6.4)
google-auth (1.1.1)
google-auth-httplib2 (0.0.2)
google-auth-oauthlib (0.1.1)
google-cloud-bigquery (0.25.0)
google-cloud-core (0.25.0)
googleapis-common-protos (1.5.3)
httplib2 (0.10.3)
idna (2.6)
numpy (1.13.3)
oauth2client (4.1.2)
oauthlib (2.0.4)
pandas (0.20.3)
pip (9.0.1)
protobuf (3.4.0)
py (1.4.34)
pyasn1 (0.3.6)
pyasn1-modules (0.1.4)
pytest (3.2.2)
python-dateutil (2.6.0)
pytz (2017.2)
requests (2.18.4)
requests-oauthlib (0.8.0)
rsa (3.4.2)
setuptools (36.5.0)
six (1.11.0)
uritemplate (3.0.0)
urllib3 (1.22)
uuid (1.30)
```

requirements.txt
jasonqng: @MaximilianR Let me see if I can figure out how to unbreak it for 0.26. If so, I'll catch any exceptions and handle. But otherwise, hopefully there is enough value to be gained that users would be willing to roll back to 0.25, if not in their daily usage, then at least in a dedicated environment when they want to use pandas-gbq in its current state, at least until we figure out a way to support future stable versions of google-cloud-bigquery

pandas_gbq/gbq.py
```diff
@@ -956,7 +956,10 @@ def _wait_for_job(job):
             bytes_billed = int(query_job._properties["statistics"]["query"].get("totalBytesProcessed", 0))
             bytes_processed = int(query_job._properties["statistics"]["query"].get("totalBytesBilled", 0))
             print("Total bytes billed (processed): %s (%s)" % (sizeof_fmt(bytes_billed),sizeof_fmt(bytes_processed)))
-    query_results = query_job.results()
+    try:
+        query_results = query_job.results()
+    except:
+        query_results = query_job.result()
```
tswast: If we require 0.26 or later, we could always use `.result()` and remove `_wait_for_job` logic from here.

requirements.txt
jasonqng: @MaximilianR Figured out why it broke. Works now for latest (0.26.0).

MaximilianR:
@jasonqng thanks, and to confirm, you're running `pytest pandas-gbq` in the root path of this repo?

MaximilianR:
@jasonqng I can't work out where `pytest_bigquery_yaml_test` comes from - there are literally [zero results](https://www.google.com/search?q=%22pytest_bigquery_yaml_test%22&oq=%22) for it from online. Any chance it's a faulty pytest plugin that might be on your system?

pandas_gbq/gbq.py
jasonqng: Hm, doc still says that the project will be inferred from environment (https://googlecloudplatform.github.io/google-cloud-python/stable/bigquery/client), but I don't think it does anymore in the latest version. Thus, project_id is now required again, which is incredibly annoying as most of the time, you're probably ok with the query being associated with the same project. Thoughts on whether we should allow the user to specify a default project env variable or other method? (`~/.bigqueryrc` looks like it can hold your default project_id, but I don't know if that's deprecated and/or a command-line only implementation: https://cloud.google.com/bigquery/bq-command-line-tool)

@MaximilianR @jreback

pandas_gbq/gbq.py
jasonqng: @tswast Latest commit uses existing GbqConnector to generate appropriate credentials. Let know if anything inappropriate with the way I've implemented.

jasonqng:
@MaximilianR Oof, yeah, maybe it came from another testing suite. I blew away the environment and started from scratch again and now it's working! Fixing the tests now... feeling like this is getting close...

jasonqng:
@jreback @tswast @parthea @MaximilianR:
After some tweaks in tests to account for new behavior, first 61 tests pass!

```
(test) ➜  pandas-gbq git:(read-gbq-google-cloud-library) ✗ pytest
=================================== test session starts ===================================
platform darwin -- Python 2.7.13, pytest-3.2.2, py-1.4.34, pluggy-0.4.0
rootdir: /Users/jasonng/Projects/pandas-gbq, inifile:
collected 61 items

pandas_gbq/tests/test_gbq.py .............................................................

=============================== 61 passed in 71.88 seconds ================================
```

Testing the last 32 generate a number of errors on my end because of a permissions issue I couldn't solve with my service account key being unable to create tables, but since I didn't touch any of the upload code, I assume this will pass if any of you all run it. Let me know if any issues still to be resolved. Look forward to any feedback.

MaximilianR:
Great! I think you need to add `google-auth` as a test dependency: https://travis-ci.org/pydata/pandas-gbq/jobs/284273205#L1142

setup.py
```diff
@@ -24,7 +24,7 @@ def readme():
     'google-auth>=1.0.0',
     'google-auth-httplib2>=0.0.1',
     'google-auth-oauthlib>=0.0.1',
-    'google-cloud-bigquery>=0.25.0,<=0.26.0',
+    'google-cloud-bigquery==0.26.0',
```
MaximilianR: I have a strong but proprietary (i.e. not everyone agrees with me) that hard pins shouldn't be in `setup.py`. If they're here then this will downgrade to `0.26.0` whenever this library is installed. 

It's OK to have them in `requirements.txt` because that doesn't force behavior.

jasonqng:
@MaximilianR Couldn't for the life of me figure out why the conda test environment kept failing with the import. `google-auth` is still in the conda requirements, I didn't touch it, only tried to add google-cloud-bigquery: https://github.com/pydata/pandas-gbq/pull/25/files#diff-810e48917c741855a0698790ac847a64R2. @jreback warned that conda dependency might be tricky, maybe he has thoughts?

setup.py
jasonqng: 0.26.0 is the latest so at the moment it would be a forced upgrade if anything (maybe I'll put 0.25.0 as an acceptable version back in so less users are forced to do so), but as others have mentioned, the major concern is that Google makes no guarantee about future behavior. At least until Google indicates that google-cloud-bigquery can be depended on for stability (at least with core functionality), I lean toward hard pinning here as well so a user always is never surprised by a broken product due to breaking changes in future versions. But happy to remove if others agree.

MaximilianR: Good points and I agree there's a tradeoff

@jreback thoughts / verdict?

pandas_gbq/gbq.py
```diff
@@ -956,61 +797,148 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+            query_job.use_legacy_sql = True
+        elif dialect == 'standard':
+            query_job.use_legacy_sql = False
+
+        if configuration:
+            for setting, value in configuration.items():
+                setattr(query_job, setting, value)
+        return query_job
+
+    def sync_query():
```
tswast: What are these `sync_query` and `async_query` functions doing here?

Only the `async_query` method should be used. `sync_query` was misnamed as it does still start a query job behind the scenes. If the call times out, you still need to poll for the job to be complete, just like with the "async" method, so you might as well remove it because it doesn't gain you anything.

requirements.txt
```diff
@@ -4,3 +4,5 @@ google-api-python-client
 google-auth
 google-auth-httplib2
 google-auth-oauthlib
+google-cloud-bigquery==0.26.0
+uuid
```
tswast: uuid is built-in and does not need to be added as a dependency.

setup.py
tswast: I know for a fact that `google-cloud[-bigquery]` will continue to make breaking changes, as I've been responsible for quite a few that will make it into later versions (0.28 or maybe 0.29).

Since `pandas-gbq` isn't currently getting built at HEAD to track the `google-cloud` changes, we won't be able to catch the breaking changes here before they release, so it's probably better to pin.

The alternative is to add some CI like we have to track Pandas changes at HEAD.

tswast:
Regarding Conda dependencies, I've got https://github.com/conda-forge/google-cloud-bigquery-feedstock/pull/1 open to update `google-cloud-bigquery` to 0.25.0. Will follow-up with a `0.26` and `0.27` release after that merges.

pandas_gbq/gbq.py
jasonqng: Initially didn't see in the docs a way to set timeout in async query, plus the docs (on the main page no less http://google-cloud-python.readthedocs.io/en/latest/bigquery/usage.html#querying-data-synchronous :) had sample for how to do so with sync query. I see in code that I can set it by passing `timeout` to `result()` in an async_query, but when I test it doesn't work. Can you confirm that this is the right way to specify a timeout?

MaximilianR:
@jreback could you ensure you're in agreement here? https://github.com/pydata/pandas-gbq/pull/25#discussion_r143250177

I'm at peace if ye sayeth

pandas_gbq/gbq.py
tswast: Yes, timeout on `result()` should work. If it does not, please file a bug with the reproducible error at https://github.com/GoogleCloudPlatform/google-cloud-python/issues

jasonqng: Issue raised here: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4135. Until I either get clarification or workaround, I'll leave sync query in for now in order to allow for timing out queries. If/when timeout works for async queries, I can easily strip out all the sync query code and implement timeout for async, hope that makes sense.

jasonqng:
@jreback Any thoughts on whether this PR is ok? All tests pass (after some tweaking of the tests), but some folks have raised questions. Let know if you need anything else on my end, would love to see this move forward if possible!

pandas_gbq/gbq.py
tswast: You're right. There's a bug in how the timeout is handled. I'll work to fix it in a subsequent version of `google-cloud-bigquery`, but you're right that this workaround is required.

ci/requirements-3.6-MASTER.pip
```diff
@@ -3,3 +3,4 @@ google-auth
 google-auth-httplib2
 google-auth-oauthlib
 mock
+google-cloud-bigquery==0.26.0
```
tswast: We might want to leave this unpinned so that the tests run on the latest version and we will know when there are breaking changes.

ci/requirements-2.7-0.19.2.pip
```diff
@@ -5,3 +5,4 @@ google-auth-oauthlib
 PyCrypto
 python-gflags
 mock
+google-cloud-bigquery==0.26.0
```
jreback: I would just add the dep, no need to fix a version in *every* test version. leave some of the empty so that they get the latest, maybe leave the 3.5 as is (as it already pins some other things).

pandas_gbq/gbq.py
```diff
@@ -7,11 +7,10 @@
...
 import os
 
-import numpy as np
-
 from distutils.version import StrictVersion
-from pandas import compat, DataFrame, concat
-from pandas.compat import lzip, bytes_to_str
+from pandas import compat, DataFrame, to_datetime, to_numeric
+from pandas.compat import bytes_to_str
+from google.cloud import bigquery
```
jreback: the import from google.cloud should in theory be *inside* _check_google_client_version

pandas_gbq/gbq.py
```diff
@@ -926,28 +749,52 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
-
-            configuration = {'query': {'useQueryCache': False}}
-
-        For more information see `BigQuery SQL Reference
-        <https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query>`__
+        <https://cloud.google.com/bigquery/sql-reference/>`
+    credentials: credentials object, default None (optional)
+        If generating credentials on your own, pass in. Otherwise, will attempt
+        to generate automatically
+    return_type: {'schema','list','df'}, default 'df'
```
jreback: anything new should have a versionadded tag

pandas_gbq/gbq.py
```diff
@@ -926,28 +749,52 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+        schema returns an array of SchemaField objects, which you can access
+            `from pprint import pprint
```
jreback: this is *very* confusing to have a return_type parameter. usually you should have separate functions for this.

pandas_gbq/gbq.py
```diff
@@ -926,28 +749,52 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+            [pprint(vars(field)) for field in schema]`
+        list returns a list of lists of the rows of the results; column names
+            are not included
```
jreback: why would you *ever* return a list of lists?

pandas_gbq/gbq.py
```diff
@@ -926,28 +749,52 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+        google-cloud-python/issues/2765> only some configuration settings are
+        currently implemented. You can pass them along like in the following:
+        `read_gbq(q,configuration={'allow_large_results':True,
+                                   'maximum_billing_tier':2})`
+        Example allowable settings:
+            allow_large_results, create_disposition, default_dataset,
+            destination, flatten_results, priority, use_query_cache,
+            use_legacy_sql, dry_run, write_disposition, udf_resources,
+            maximum_billing_tier, maximum_bytes_billed
+            <http://google-cloud-python.readthedocs.io/en/latest/_modules/
```
jreback: alot of this won't render. build the docs locally and you will see many sphinx errors

pandas_gbq/gbq.py
```diff
@@ -956,61 +803,148 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
-    _test_google_api_imports()
-
-    if not project_id:
-        raise TypeError("Missing required parameter: project_id")
-
     if dialect not in ('legacy', 'standard'):
         raise ValueError("'{0}' is not valid for dialect".format(dialect))
+    if configuration and any(key in configuration for key in
+                             ["query", "copy", "load", "extract"]):
+        raise ValueError("New API handles configuration settings differently")
```
jreback: add an informative message

pandas_gbq/gbq.py
```diff
@@ -956,61 +803,148 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+            for setting, value in configuration.items():
+                setattr(query_job, setting, value)
+        return query_job
+
+    def sync_query():
+        query_job = client.run_sync_query(query,
+                                          query_parameters=query_parameters)
+        query_job = _set_common_query_settings(query_job)
+        if verbose:
+            print("Query running...")
```
jreback: since you are re-writing. I would remove all prints and simply use the standard logging module, much more modular / extensible.

pandas_gbq/gbq.py
```diff
@@ -956,61 +803,148 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+                print("Cache hit.")
+            elif ("statistics" in query_job._properties and
+                    "query" in query_job._properties["statistics"]):
+                bytes_billed = int(query_job
+                                   ._properties["statistics"]["query"]
+                                   .get("totalBytesProcessed", 0))
+                bytes_processed = int(query_job
+                                      ._properties["statistics"]["query"]
+                                      .get("totalBytesBilled", 0))
+                print("Total bytes billed (processed): %s (%s)" %
```
jreback: why do you have asyc and sync? generally you would *always* do async, and just wait on it for sync.

pandas_gbq/gbq.py
```diff
@@ -956,61 +803,148 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+                  query_job.created.replace(tzinfo=None)).seconds)
+            print("Finished at %s." % datetime.now()
+                  .strftime('%Y-%m-%d %H:%M:%S'))
+
+    if return_type == 'schema':
+        return query_results.schema
+    elif return_type == 'list':
+        return rows
+
+    columns = [field.name for field in query_results.schema]
```
jreback: you have a whole module, so splitting this function would be a good idea.

pandas_gbq/gbq.py
```diff
@@ -956,61 +803,148 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
...
+    data = rows
+
+    final_df = DataFrame(data=data, columns=columns)
+
```
jreback: it seems that you are defining all of these functions *within* the read_gbq function? why is that

pandas_gbq/tests/test_gbq.py
```diff
@@ -13,10 +13,13 @@
 from pandas import compat
 
 from pandas.compat import u, range
-from pandas import NaT, DataFrame
+from pandas import NaT, DataFrame, to_datetime
 from pandas_gbq import gbq
 import pandas.util.testing as tm
 from pandas.compat.numpy import np_datetime64_compat
+from google.cloud import bigquery
```
jreback: you prob need a pytest.importorskip here

requirements.txt
```diff
@@ -4,3 +4,4 @@ google-api-python-client
 google-auth
 google-auth-httplib2
 google-auth-oauthlib
+google-cloud-bigquery==0.26.0
```
jreback: not needed

setup.py
```diff
@@ -24,6 +24,7 @@ def readme():
     'google-auth>=1.0.0',
     'google-auth-httplib2>=0.0.1',
     'google-auth-oauthlib>=0.0.1',
+    'google-cloud-bigquery==0.26.0',
```
jreback: >=

pandas_gbq/gbq.py
jasonqng: I think I see the point of doing the checks before the import of bigquery, but wouldn't putting the import of bigquery inside _check cause it to be scoped for only inside that function, and thus inaccessible to read_gbq? But feel free to correct and suggest how I should do it since I'm still a python novice, thank you!

pandas_gbq/gbq.py
jasonqng: See thread here: https://github.com/pydata/pandas-gbq/pull/25/files/61db19190cdfe690298288cf90e0d62a5b73e95d#r143248553. There's a bug in how async handles timeouts. For now, to provide backwards compatibility for timeouts, we need to do this workaround. When it is fixed, we can then clean this up and remove the sync code. Hope that makes sense, but let know if you think this should be handled differently.

setup.py
tswast: I believe the code as-written would also function on 0.27.0, so a range like `google-cloud-bigquery>=0.26.0,<0.28.0` would be better.

Note that [0.28.0 is coming soon](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4258) which includes many breaking changes, but also fixes the timeout issue mentioned above.

pandas_gbq/gbq.py
jasonqng: I'm replicating how it was written in the previous version of read_gbq, which similarly handled re-ordering and type conversion all within read_gbq. If you think this should be refactored, I can take a crack at it, but perhaps it can be left for another day? The schema/column names, I can try splitting out into another function to roughly mirror how it was done before if that makes sense.

pandas_gbq/tests/test_gbq.py
jasonqng: Sorry, I'm not that familiar with pytest. Can you provide more guidance or perhaps someone else can let know what I should do? Otherwise, I can poke around when I have time, but that might not be for a while :(

setup.py
jasonqng: Can you clarify? Change `==` to `=` causes setup to fail due to parse error.

jasonqng: @tswast 👍 Switched to your suggestion

requirements.txt
jasonqng: Should it not be pinned in requirements? See discussion here: https://github.com/pydata/pandas-gbq/pull/25#pullrequestreview-67723587

pandas_gbq/gbq.py
jasonqng: Can we leave that for another day since I don't know the convention and there's no model to follow here?

pandas_gbq/gbq.py
jasonqng: I can research, but perhaps someone can save me time and point to a resource since I'm not familiar with what does and doesn't render documentation-wise.

ci/requirements-2.7-0.19.2.pip
jasonqng: Good point, updated.

pandas_gbq/gbq.py
jasonqng: :) I can imagine some use cases, but I hear ya, I've removed.

pandas_gbq/gbq.py
jasonqng: Was originally meant to handle the verify schema test, but I've re-written so no need for return type anymore. You can access the schema by using the `run_query` function now.

pandas_gbq/gbq.py
jasonqng: 👍

pandas_gbq/gbq.py
jasonqng: 👍 I split all the schema work out into a `get_columns_schema()` function.

jasonqng:
@jreback Took some steps at making it more modular and did my best to address your other issues. Is there anything still holding this back? Happy to do the essentials (is fixing the rendering of the documentation essential?), but anything else that is nice to have can perhaps wait for another day/for another person. Thanks again for your thorough review and helpful comments.

All tests still pass:
```
============================= test session starts ==============================
platform darwin -- Python 2.7.13, pytest-3.2.2, py-1.4.34, pluggy-0.4.0
rootdir: /Users/jasonng/Projects/pandas-gbq, inifile:
collected 61 items

test_gbq.py .............................................................

========================== 61 passed in 87.01 seconds ==========================
```

pandas_gbq/gbq.py
jasonqng: If I run `help(gbq.read_gbq)` or do shift+tab in Jupyter Notebook, the documentation seems to load fine. If you have an example of what's broken and the circumstance, happy to try and fix. Thank you!

tswast: Link rendering syntax: http://www.sphinx-doc.org/en/stable/rest.html#external-links

jasonqng: Thanks, I went through and fixed hyperlinks and cleaned it up a bit according to that doc. Tried to build locally from the existing docs folder. Console output appeared to show that Sphinx update ran successfully, but when I actually load the api html page, it still shows the existing documentation. I even deleted the html file, but it just rebuilt them with the existing documentation. So I can't actually verify that it works (perhaps someone else can?), but just eyeballing the docstring, it should be ok now.

pandas_gbq/tests/test_gbq.py
MaximilianR: This becomes a hard dependency (right @jasonqng ?), so no skipping

MaximilianR:
> @jreback Took some steps at making it more modular and did my best to address your other issues. Is there anything still holding this back? Happy to do the essentials (is fixing the rendering of the documentation essential?), but anything else that is nice to have can perhaps wait for another day/for another person. 

I think this is right - let's try and get this in, as long as it meets minimum requirements. We can push things like the log / print off to another PR (I'm in debt on this one, for example). 

What are the final questions? @jreback ? https://github.com/pydata/pandas-gbq/pull/25#pullrequestreview-67723587 needs a verdict from you, but otherwise are you happy to proceed?

pandas_gbq/gbq.py
jreback: you can't have the imports like this as I stated above. Then the entire module will simply be an import error (and not a friendly one). The entire point is to provide a nice error message.

pandas_gbq/gbq.py
```diff
@@ -837,68 +681,172 @@ def _get_credentials_file():
...
+    -------
+    Tuple
+        rows : list of lists
+        columns: list of strings
+        schema: dictionary
+            Has the following keys: name, field_type, mode, fields, description
+    """
+
+    def _wait_for_job(job):
+        while True:
```
jreback: so this nesting of functions is making this really really hard to understand. Please separate everything out to a top level function. You will want to create a separate module (maybe call it query.py or something), maybe need a utils.py or whatever. You want to be able to see at a glance what is actually happening, and abstract away the details.

pandas_gbq/gbq.py
jreback: yes this is ok

pandas_gbq/gbq.py
jreback: see my point above, this should be abstracted away, again to a query.py or something to make this obvious to the reader.

pandas_gbq/gbq.py
```diff
@@ -837,68 +681,172 @@ def _get_credentials_file():
...
+                                   .get("totalBytesProcessed", 0))
+                bytes_processed = int(query_job
+                                      ._properties["statistics"]["query"]
+                                      .get("totalBytesBilled", 0))
+                print("Total bytes billed (processed): %s (%s)" %
+                      (sizeof_fmt(bytes_billed), sizeof_fmt(bytes_processed)))
+            print("\nRetrieving results...")
+        return query_results, query_job
+
+    def get_columns_schema(query_results):
```
jreback: this is still 1 really really long function.

pandas_gbq/gbq.py
```diff
@@ -837,68 +681,172 @@ def _get_credentials_file():
...
+    Returns
+    -------
+    Tuple
+        rows : list of lists
+        columns: list of strings
+        schema: dictionary
+            Has the following keys: name, field_type, mode, fields, description
+    """
+
+    def _wait_for_job(job):
```
tswast: This function is unnecessary as of [google-cloud-bigquery version 0.26.0](https://github.com/GoogleCloudPlatform/google-cloud-python/releases/tag/bigquery-0.26.0), which implements the `concurrent.futures` API for all job types.

pandas_gbq/gbq.py
tswast: Fixed in [0.28.0](https://github.com/GoogleCloudPlatform/google-cloud-python/releases/tag/bigquery-0.28.0) (Though that includes a bunch of other breaking changes, so I understand if you don't want to update this PR. Sorry about that!)

pandas_gbq/gbq.py
```diff
@@ -837,68 +681,172 @@ def _get_credentials_file():
...
-    page_array = np.zeros((len(rows),), dtype=lzip(col_names, col_dtypes))
-    for row_num, raw_row in enumerate(rows):
-        entries = raw_row.get('f', [])
-        for col_num, field_type in enumerate(col_types):
-            field_value = _parse_entry(entries[col_num].get('v', ''),
-                                       field_type)
-            page_array[row_num][col_num] = field_value
 
-    return DataFrame(page_array, columns=col_names)
+def run_query(query, client, dialect='legacy', query_parameters=(),
```
tswast: Nit: this PR would be easier to review if the `run_query()` method didn't move (stayed defined just before `load_data()`)

pandas_gbq/gbq.py
```diff
@@ -956,63 +931,79 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
 
     """
 
+    def _create_df(rows, columns, schema, index_col, col_order):
```
tswast: Once https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4354 goes in (adds `to_dataframe()` method to the `google-cloud-bigquery` library), the `_create_df` function may be unnecessary.

pandas_gbq/gbq.py
jasonqng: Perhaps I'm misreading your comment? See my response about the imported module being no longer in scope if I import it in `_check_google_client_version()`. Doing so will give me a `NameError: global name 'bigquery' is not defined` error going forward. Perhaps I could import after `_test_google_api_imports()` inside the `read_gbq()` function? Would that resolve the issue for you?

pandas_gbq/gbq.py
jasonqng: As per @tswast, this function is now unnecessary. Tested and confirmed. Will delete.

pandas_gbq/gbq.py
jasonqng: Yeah, going to try and get this working version merged before having to deal with new breaking changes, thanks for the update! :)

pandas_gbq/gbq.py
jasonqng: Good idea, moved this back to GbqConnector and moved away some more of the plumbing in read_gbq to GbqConnector as well.

pandas_gbq/gbq.py
jasonqng: I realize that it may look a little long, but it's no longer nor any more complex than the previous run_query function (it's actually 30 lines shorter). And if google-cloud handled timeouts in async queries correctly, this would be another 30 lines shorter and two functions less. I'm not sure why or how this function would be split up even more as it follows the same logic as the original run_query: you pass it a query and it handles all the client work and returns the data along with the schema.

jasonqng:
@jreback I appreciate your comments and encouragement to abstract away more of the plumbing and I've taken a number of efforts to do so with your help. I agree that readable, modular code is always ideal. 

However, I'm hopeful that you agree that this PR and the way I've re-written the read_gbq() and run_query() functions is no less readable or modular than the existing code. I could spend more time moving various parts of the run_query() code to a separate module, but at this point, it feels like shuffling code around and doesn't actually advance the point of this PR, which is to shift away from the legacy big query python library over to the actively maintained google-cloud package, with its out of the box support for arrays and structs included, which was why I originally started working on this.

Could you point out specific things I should correct before this is mergable? I'm happy to fix any showstopping issues with the time I can spare, but if your previous suggestions were primarily about making the code a bit more readable, could we perhaps save that for a future iteration? Or perhaps I should hand this off to someone else who knows what to do to push this over the finish line? Comments from @tswast @parthea @maxim-lian about what final things I need to do also appreciated!

Thanks again for the very thoughtful suggestions and I'll keep them in mind when working on any future projects!

pandas_gbq/gbq.py
jasonqng: Sounds good, look forward to it!

pandas_gbq/gbq.py
jasonqng: Removed

jasonqng:
(FYI, tweaked the tests in earlier commit to account for the fact that credentials for the client are handled directly inside GbqConnector, and there's no need to pass them from read_gbq. Verified that all tests still pass.)

pandas_gbq/gbq.py
jasonqng: FYI, moved the import to inside GbqConnector since that's the only place it's needed (to generate the appropriate client object). read_gbq performs a `_test_google_api_imports()` at the very start, so if there are import issues, `_test_google_api_imports()` will provide an error message.

tswast:
Hey @jasonqng, thanks for your hard work on this. I've made some changes to this PR, just so you aren't caught off-guard too much.

- I noticed there were still uses of the API client library, so I converted the rest of methods to use google-cloud-bigquery
- Removes references to google-api-client-library and httplib2.
- Updates PR to not make any surface-level changes to the API, only
   swaps out the dependencies.
- Updates PR to use latest version of google-cloud-bigquery (0.28.0).

There's one test failure right now because the cloud client library requires the "mode" field for schemas, which wasn't previously provided.

jasonqng:
@tswast SGTM! Thanks for the updates. I'm on board with whatever you and @jreback think is best to get this finalized. Let know if you need anything else on my end!

jreback:
does this reference any current issue? (e.g. that this will close)

jreback:
also after this is merged, will need to do a PR to main pandas to update the dependencies on the CI (to use the new ones).

tswast:
There wasn't an issue for this change, so I just created https://github.com/pydata/pandas-gbq/issues/93.

I'll look more into the CI failures, docs updates, and upstream pandas changes required today.

jasonqng:
This PR will also close these two issues (which were the reason I started down this rabbithole :)

https://github.com/pydata/pandas-gbq/issues/22
https://github.com/pydata/pandas-gbq/issues/23

It may also close this one?
https://github.com/pydata/pandas-gbq/issues/69

jasonqng:
@tswast I pushed a quick commit that should hopefully resolve the CI failure (feel free to squash). I'll leave the docs and any upstream changes to you, much appreciate all your help!

tswast:
Thanks!

tswast:
Upstream docs changes at https://github.com/pandas-dev/pandas/pull/18548

tswast:
Re "ci on upstream pandas": The two files that have `pandas-gbq` listed as a dependency are

- https://github.com/pandas-dev/pandas/blob/master/ci/requirements-2.7.pip
- https://github.com/pandas-dev/pandas/blob/master/ci/requirements-3.6_DOC.sh

Neither explicitly adds any of `pandas-gbq`'s dependencies.

tswast:
I'm getting one test failure after merging changes from master. Looking into it now.

tswast:
Oh, I guess it was just the X-skip that's already there.

```
$ pytest pandas_gbq
============================================= test session starts =============================================
platform darwin -- Python 3.6.1, pytest-3.3.0, py-1.5.2, pluggy-0.6.0
rootdir: /Users/tswast/src/pandas/pandas-gbq, inifile:
collected 95 items

pandas_gbq/tests/test_gbq.py .......s...........................................................X.s.... [ 77%]
.....................                                                                                   [100%]

============================== 92 passed, 2 skipped, 1 xpassed in 601.79 seconds ==============================
```

tswast:
@jreback Could you take a look at the docs changes I made? I've documented both the dependencies for 0.3.0 and what they were before (with notes on how they've changed)

pandas_gbq/gbq.py
```diff
@@ -781,6 +656,14 @@ def verify_schema(self, dataset_id, table_id, schema):
                                key=lambda x: x['name'])
         fields_local = sorted(schema['fields'], key=lambda x: x['name'])
 
+        # Ignore mode when comparing schemas.
+        for field in fields_local:
+            if 'mode' in field:
```
maxim-lian: not worth changing, but this could be marginally simpler as `field.pop('mode', None)`

pandas_gbq/gbq.py
```diff
@@ -847,44 +738,31 @@ def _parse_data(schema, rows):
     fields = schema['fields']
     col_types = [field['type'] for field in fields]
     col_names = [str(field['name']) for field in fields]
-    col_dtypes = [dtype_map.get(field['type'], object) for field in fields]
+    col_dtypes = [
+        dtype_map.get(field['type'].upper(), object)
+        for field in fields
+    ]
+    print(fields)
```
maxim-lian: Do we want to print all the fields here?

If we do: elsewhere, this uses `self._print`, which can can transition to logging at some point.

pandas_gbq/gbq.py
```diff
@@ -577,141 +500,97 @@ def run_query(self, query, **kwargs):
...
+                body = StringIO('{}\n'.format('\n'.join(rows)))
 
                 try:
-                    response = self.service.tabledata().insertAll(
-                        projectId=self.project_id,
-                        datasetId=dataset_id,
-                        tableId=table_id,
-                        body=body).execute()
-                except HttpError as ex:
+                    self.client.load_table_from_file(
```
maxim-lian: This is so much better than the existing method

pandas_gbq/gbq.py
```diff
@@ -577,141 +500,97 @@ def run_query(self, query, **kwargs):
...
         for index, row in dataframe.reset_index(drop=True).iterrows():
-            row_dict = dict()
-            row_dict['json'] = json.loads(row.to_json(force_ascii=False,
-                                                      date_unit='s',
-                                                      date_format='iso'))
-            row_dict['insertId'] = job_id + str(index)
-            rows.append(row_dict)
+            row_json = row.to_json(
+                force_ascii=False, date_unit='s', date_format='iso')
+            rows.append(row_json)
```
maxim-lian: This isn't worse than the last version, but this would be *much* faster if `.to_json` was called on the whole table rather than each row, iterating in python

CSV might be even faster given the reduced space (and pandas can't use nesting or structs anyway). But potentially wait until parquet is GA

pandas_gbq/gbq.py
```diff
@@ -847,44 +738,31 @@ def _parse_data(schema, rows):
...
+    print(fields)
     page_array = np.zeros((len(rows),), dtype=lzip(col_names, col_dtypes))
-    for row_num, raw_row in enumerate(rows):
-        entries = raw_row.get('f', [])
-        for col_num, field_type in enumerate(col_types):
-            field_value = _parse_entry(entries[col_num].get('v', ''),
-                                       field_type)
+    for row_num, entries in enumerate(rows):
+        for col_num in range(len(col_types)):
+            field_value = entries[col_num]
```
maxim-lian: I don't think I'd realized we were looping over all the values in python before. This explains a lot of why exporting a query to a file on GCS and then reading from that file is an order of magnitude faster. 

If we could pass `rows` directly into `DataFrame` that would be much faster, but I'm not sure if that's possible

maxim-lian: Because this is being called so many times, you may even get a small speed up from eliminating assigning to `field_value`

(but these are all things that are either the same or better than the existing version)

pandas_gbq/gbq.py
```diff
@@ -577,141 +500,97 @@ def run_query(self, query, **kwargs):
...
+                bytes_processed = query_reply.total_bytes_processed or 0
+                bytes_billed = query_reply.total_bytes_billed or 0
+                self._print('Query done.\nProcessed: {} Billed: {}'.format(
+                    self.sizeof_fmt(bytes_processed),
+                    self.sizeof_fmt(bytes_billed)))
                 self._print('Standard price: ${:,.2f} USD\n'.format(
-                    bytes_processed * self.query_price_for_TB))
+                    bytes_billed * self.query_price_for_TB))
 
             self._print('Retrieving results...')
```
maxim-lian: Presumably this is never going to be hit because the prior part is blocking?

maxim-lian:
I took a proper read through.

I think this strictly dominates the existing version. There are a couple of extremely small tweaks that we can do in a follow-up if not now.

There are also some areas for huge speed-ups - IIUC the code is currently running through each value in python atm. Internally, we've built a function for exporting to a file to GCS and loading that in, which works much better for > 1-2m rows. We can do a PR for that if people are interested, in addition to speeding up the current path.

pandas_gbq/gbq.py
tswast: I'd prefer to keep the current behavior for now and do a subsequent PR for any changes like this for performance improvements. I've filed https://github.com/pydata/pandas-gbq/issues/96 to track the work for speeding up encoding for the `to_gbq()` method.

pandas_gbq/gbq.py
tswast: Yeah, it's technically a change in behavior (kicks off a load job instead of using the streaming API), but I think the change is small enough to be worth it. Load jobs should be much more reliable for the use case of this library.

pandas_gbq/gbq.py
tswast: Yes and no. I think it indeed is less relevant, but actually fetching the rows and constructing the dataframe will take non-zero time, especially for larger result sets.

pandas_gbq/gbq.py
tswast: Oops. Good catch. I added that to debug the "ignore mode when comparing schemas" logic. Removed.

pandas_gbq/gbq.py
tswast: I've filed https://github.com/pydata/pandas-gbq/issues/97 to track improving the performance in the read case.

pandas_gbq/gbq.py
maxim-lian: 100% re doing that separately

maxim-lian:
@jreback I know there's lots going on in pandas, but would be super if you could take a glance at this. A few follow-ups are dependent on this merging.

Thanks v much

jreback:
sure will look

docs/source/changelog.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+0.3.0 / 2017-??-??
+------------------
+
+- Use the `google-cloud-bigquery <https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/usage.html>`__ library for API calls instead of ``google-api-client`` and ``httplib2``. (:issue:`93`)
```
jreback: needs a more comprehensive note here. show what you used to import / depend on, and what it should be now.

jreback: you can link to install.rst as well

pandas_gbq/gbq.py
```diff
@@ -181,14 +158,6 @@ class QueryTimeout(ValueError):
     pass
 
 
-class StreamingInsertError(ValueError):
```
jreback: mention that this is eliminated in whatsnew

maxim-lian:
Congrats @jasonqng & @tswast !

tswast:
Thanks! A couple things to clean up before we make a release. I'd like to add a couple tests for some of the other issues we think this PR might address.

Plus I'm not sure it makes sense to do a release right before the holidays.

jasonqng:
TY @tswast for carrying this across the finish line!!! I learned so much watching you and @jreback mold this. Can't wait to stop having to rely on my janky fork and install the new version!

PR #40, opened 2017-05-18
COMPAT: drop 3.4 support

jreback:
TST: drop 3.4, add 3.6/0.20.1 to config matrix


jreback:
@parthea this seems to break some tests. any idea?

parthea:
At first glance I'm not sure about the root cause of the broken tests but I will take a look tonight and fix them.

tswast:
I encountered similar test failures. Seems like the root cause is eventual consistency in the table and dataset listing on the BigQuery backend.

Changing `clean_gbq_environment` and removing the orphaned datasets from the project seems to have improved the stability in my tests.

```
def clean_gbq_environment(dataset_prefix, private_key=None):
    dataset = gbq._Dataset(_get_project_id(), private_key=private_key)
    all_datasets = dataset.datasets()

    retry = 3
    while retry > 0:
        try:
            retry = retry - 1
            for i in range(1, 10):
                dataset_id = dataset_prefix + str(i)
                if dataset_id in all_datasets:
                    table = gbq._Table(_get_project_id(), dataset_id,
                                       private_key=private_key)
                    all_tables = dataset.tables(dataset_id)
                    for table_id in all_tables:
                        try:
                            table.delete(table_id)
                        except gbq.NotFoundException, e:
                            pass

                    dataset.delete(dataset_id)
            retry = 0
        except gbq.GenericGBQException as ex:
            # Build in retry logic to work around the following errors :
            # An internal error occurred and the request could not be...
            # Dataset ... is still in use
            error_message = str(ex).lower()
            if ('an internal error occurred' in error_message or
                'still in use' in error_message) and retry > 0:
                sleep(randint(1, 10))
            else:
                raise ex

MaximilianR:
I think there's a change on the backend that may have caused this - all our internal integration tests with BQ have broken. 

These are two SO questions that I posted over the last 24 hours: 
http://stackoverflow.com/questions/44053351/weird-behavior-with-from-list-tables-in-google-bigquery?noredirect=1&lq=1
http://stackoverflow.com/questions/44030182/table-consistency-in-google-bigquery/44030768?noredirect=1#comment75097001_44030768

tswast:
I'll poke the BigQuery engineers to see if there's something they could roll back. The fact that dataset listing can fail with the provided token is really annoying.

PR #39, opened 2017-05-18
BUG: oauth2client deprecated, use google-auth instead.

tswast:
Remove the use of oauth2client and use google-auth library, instead.

Rather than check for multiple versions of the libraries, use the
setup.py to specify compatible versions. I believe this is safe since
Pandas checks for the pandas_gbq package.

Since google-auth does not use the argparse module to override user
authentication flow settings, add a parameter to choose between the web
and console flow.

Closes https://github.com/pydata/pandas-gbq/issues/37.

PR #58, opened 2017-06-11
TST: Fix broken tests failing with 'NoneType' object is not iterable / module 'pandas.util.testing' has no attribute 'assertRaises'

### LINKED ISSUE #57: TST: failing tests
seems our tests have been failing for a bit: https://travis-ci.org/pydata/pandas-gbq/builds/233574394

any ideas @parthea ?

parthea:
Fixes #57. This 2 line fix was originally part of @tswast's PR #39 but I'm going to transfer it here so that it's easier to capture the specific fix. PR #39 also has other reliability improvements to the integration testing which resolve intermittent failures. The changes in this PR fix the consistent failure `TypeError: 'NoneType' object is not iterable`.

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -1055,6 +1055,9 @@ class _Dataset(GbqConnector):
                 dataset_response = list_dataset_response.get('datasets')
                 next_page_token = list_dataset_response.get('nextPageToken')
 
+                if dataset_response is None:
+                    dataset_response = []
+
                 for row_num, raw_row in enumerate(dataset_response):
                     dataset_list.append(
                         raw_row['datasetReference']['datasetId'])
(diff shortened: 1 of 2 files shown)


parthea:
All tests are passing now with the changes in this PR.
https://travis-ci.org/parthea/pandas-gbq/builds/241645611

parthea:
Thanks @tswast !

jreback:
can u add a release note as well

parthea:
My initial thought is that a release note entry isn't needed because the affected function  `_Dataset(<project_id>).datasets()` at https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/gbq.py#L1030 is not called in `gbq.py`. It is only called in integration testing. A release entry is needed if we think someone may be using this specific undocumented function.

jreback:
@parthea ahh I c. that is all fine. glad tests are passing. thanks!

Issue #7, opened 2017-02-08
Bulk upload


maxim-lian:
ref: https://github.com/pandas-dev/pandas/issues/14670

R does this well, and it makes a material performance impact

MaximilianR:
I've got this working internally, by writing to a CSV and then uploading that.

I've used the `google.cloud.bigquery` library though, which this library doesn't seem to use. Is there an explicit reason for this? The credentials are handled a bit differently.

jreback:
you could certainly add that as a dep. looks like that is a refactor of some older libs?

MaximilianR:
OK

jreback:
yeah looks like these are the 'new' ones: https://cloud.google.com/bigquery/docs/reference/libraries#client-libraries-usage-python

tswast:
Fixed by https://github.com/pydata/pandas-gbq/pull/25. I'll update the changelog to link to this.

melissachang:
What version of pandas will this be in? How do I find out?

tswast:
@melissachang It is now in the released version of `pandas-gbq` 0.3.0. https://github.com/pydata/pandas-gbq/releases/tag/0.3.0

PR #60, opened 2017-06-22
ENH: Supporting REPEATED schema for list types

txomon:
Basically, at the moment if there are lists inside a column, we have no way to put that data in GBQ. This enables REPEAT mode in a simple way.

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -1079,8 +1079,11 @@ def _generate_bq_schema(df, default_type='STRING'):
 
     fields = []
     for column_name, dtype in df.dtypes.iteritems():
-        fields.append({'name': column_name,
-                       'type': type_mapping.get(dtype.kind, default_type)})
+        definition = {'name': column_name,
+                      'type': type_mapping.get(dtype.kind, default_type)}
+        if isinstance(getattr(df, column_name)[0], list):
+            definition['mode'] = 'REPEATED'
+        fields.append(definition)
 
     return {'fields': fields}
 
--- a/pandas_gbq/tests/test_gbq.py
+++ b/pandas_gbq/tests/test_gbq.py
@@ -1092,11 +1092,14 @@ class TestToGBQIntegrationWithServiceAccountKeyPath(object):
         assert schema == test_schema
 
     def test_generate_repeated_schema(self):
-        df = DataFrame([{'A': [1, 2, 3], 'B': ['1', '2']}, {'A': [3, 5], 'B': ['8', '3']}])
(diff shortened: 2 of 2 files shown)


jreback:
tests!

txomon:
I don't get what if failing, the test is there but doesn't look like it's being executed.... 🤔

MaximilianR:
FWIW it's not 'pandantic' to put lists in a dataframe, and IIRC if they're on an index you'll face issues with a lot of pandas functionality

We approach this with either:
a) separate dataframes uploaded to BQ, and then run a BQ query to merge them into a nested format in a single table
b) stack / unstack, with a BQ query to merge columns

txomon:
@MaximilianR I guessed it is not, but is something we use, thankfully not too often, but still need. 

 1. Should we put a warning or something to warn about a list schema being generated?
 2. Can you tell me why the test isn't being executed?

MaximilianR:
What makes you think it's not being executed?

parthea:
Thanks @txomon! I think #25 may help us provide support for BigQuery repeated fields. I'd like to keep this PR open just in case #25 doesn't provide this functionality. 

Regarding the integration tests being skipped, some integration tests will be skipped if a BigQuery project id is not set. I think the tests should fail if a project id is not set. I've created #72 to track this improvement.

Follow these steps to run the BigQuery integration tests on Travis:
https://pandas-gbq.readthedocs.io/en/latest/contributing.html#running-google-bigquery-integration-tests

tswast:
Yeah, #25 may help. This is also something to make sure we handle when we switch from streaming writes to bulk upload https://github.com/pydata/pandas-gbq/issues/7.

maxim-lian:
Anyone know whether this is still an issue?

tswast:
A similar issue was fixed for `read_gbq` in https://github.com/pydata/pandas-gbq/pull/134.

I think this PR would probably be useful if we wanted to support list types in `to_gbq`, but the fix would be more involved. (Would need to switch to a different file type than CSV for load jobs, maybe Parquet?)

maxim-lian:
Right, and I think then pandas is not the right tool to be using for writing nested items - lists are badly supported if at all (I'm doubtful that it can write to nested parquet files, for example, so we'd have to write that part too).

We've thought about something that would go from `xarray` (i.e. 3+ dimensions) to a nested format to BQ, but even then there are difficulties (which dimension becomes nested? How can you create the JSON / dict in a memory-efficient way in python?)

I would vote to close, but v open if there are any creative ideas

max-sixty:
Closing as stale, but open to a solution for writing repeated fields

Issue #63, opened 2017-06-28
DOC: Add conda install steps for google-auth*


parthea:
From #62, we need to include steps in the pandas-gbq documentation on how to install the following:

```
google-auth
```
```
google-auth-httplib2
```
```
google-auth-oauthlib
```

tswast:
I believe this should automatically be installed based on the dependencies listed at https://github.com/conda-forge/pandas-gbq-feedstock/blob/master/recipe/meta.yaml after we install the next version, but we should show this for when folks install the library in development mode: `pip install -e .` along with `conda develop`.

Issue #20, opened 2017-03-04
Improve code coverage when gbq credentials are not provided


parthea:
By default, the majority of pandas-gbq unit tests are skipped on forks of pandas-gbq because Google BigQuery integration testing requires credentials. This causes code coverage to appear low on pull requests from forks. One potential way to increase the code coverage reported in pull requests from forks is to use `mock` to stub external requests.
https://docs.python.org/3.6/library/unittest.mock.html

jreback:
yeah I realized that now :>

we have a very low check on the PR's anyhow

maxim-lian:
@tswast is there any interface Google provides that would allow testing against BQ without an account / with a limited-google-provided-account? At the moment people can't run tests without a google account (and given https://github.com/pydata/pandas-gbq/issues/109, without downloading a json file and putting it in a specific path)

tswast:
A local emulator would be a good way to provide such a use-case. Google has them for some services like Bigtable and Cloud Pub/Sub, but there isn't one for BigQuery.

maxim-lian:
OK thanks. 
This might be a long-shot, but do you think Google might be willing to provide some keys we could put in the repository so, for example, any PR pushed to Travis would test?

tswast:
I believe this project already has some secret keys used to build on merge.

Some CI systems allow accessing secret keys after a PR from a fork has been manually reviewed by adding a GitHub tag/label to the PR. I don't believe Travis has this feature, though.

maxim-lian:
ok thx

max-sixty:
Unfortunately I don't think there's a BigQuery mocking library like there is with some other Google products. I agree that would radically improve the testing iteration. 

But until then, I propose to close this as out-of-scope; unless people think I'm overestimating how difficult it would be to build the mocks

tswast:
Yeah, no BigQuery emulator. It does make unit testing annoying / infeasible.

PR #64, opened 2017-06-29
DOC: Add development installation instructions

tswast:
Includes conda install steps for dependencies.

Closes https://github.com/pydata/pandas-gbq/issues/63.

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -139,6 +139,52 @@ request.  If you have uncommitted changes, you will need to ``stash`` them prior
 to updating.  This will effectively store your changes and they can be reapplied
 after updating.
 
+Install in Development Mode
+---------------------------
+
+It's helpful to install pandas-gbq in development mode so that you can
+use the library without reinstalling the package after every change.
+
+Conda
+~~~~~
+
+Create a new conda environment and install the necessary dependencies
+
+.. code-block:: shell
+
+   $ conda create -n my-env --channel conda-forge  \
+         pandas \
+         google-auth-oauthlib \
+         google-api-python-client \
+         google-auth-httplib2
+   $ source activate my-env
+
+Install pandas-gbq in development mode
+
+.. code-block:: shell
+
+    $ python setup.py develop
+
+Pip & virtualenv
+~~~~~~~~~~~~~~~~
+
+*Skip this section if you already followed the conda instructions.*
+
+Create a new `virtual
+environment <https://virtualenv.pypa.io/en/stable/userguide/>`__.
+
+.. code-block:: shell
+
+    $ virtualenv env
+    $ source env/bin/activate
+
+You can install pandas-gbq and its dependencies in `development mode via
+pip <https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs>`__.
+
+.. code-block:: shell
+
+    $ pip install -e .
+
 Contributing to the code base
 =============================
 


docs/source/install.rst
```diff
@@ -32,6 +32,60 @@ Install from Source
...
+
+.. code-block:: shell
+
+    $ conda develop .
+
+Install the necessary dependencies
+
+.. code-block:: shell
+
+    $ conda install --channel conda-forge \
```
jreback: so the point of this is you *don't* have to specify secondary deps, they will simply be installed via the dependency graph. it looks like you install

``conda create -n foo --dry google-auth-oauthlib google-api-python-client pandas google-auth-httplib2``

though I am suprised that ``google-auth`` is actually a dep of the ``google-auth-oauthlib`` (and httplib2), you usually don't do it this way, rather installing ``google-auth`` should install all libs under its tree, so this is a bit odd.

docs/source/install.rst
```diff
@@ -32,6 +32,60 @@ Install from Source
...
+.. code-block:: shell
+
+   $ conda create --no-deps -n pandas-gbq
+   $ source activate pandas-gbq
+
+Install pandas-gbq in development mode
+
+.. code-block:: shell
+
+    $ conda develop .
```
jreback: we use ``python setup.py develop`` rather than ``conda develop``

docs/source/install.rst
```diff
@@ -32,6 +32,60 @@ Install from Source
...
+
+.. code-block:: shell
+
+    $ conda install conda-build
+
+Optionally, create a new conda environment.
+
+.. code-block:: shell
+
+   $ conda create --no-deps -n pandas-gbq
```
jreback: you can do this all in one go, see below

docs/source/install.rst
```diff
@@ -32,6 +32,60 @@ Install from Source
...
+
+Conda
+~~~~~
+
+To install packages in development mode using conda, first
+install the conda build tools package in your root conda environment.
+
+.. code-block:: shell
+
+    $ conda install conda-build
```
jreback: you don't need conda-build

docs/source/install.rst
```diff
@@ -32,6 +32,60 @@ Install from Source
...
+
+Optionally, create a new conda environment.
```
jreback: this is not optional, you *always* do this.

docs/source/install.rst
```diff
@@ -32,6 +32,60 @@ Install from Source
...
+    $ cd pandas-gbq
+
+Pip
+~~~
+
+To install pandas-gbq and its dependencies in development mode via pip
+
+.. code-block:: shell
+
+    $ pip install -e .
```
jreback: we generally don't use pip to install things when we have proper conda packages. sure you can, but its not the first thing you do.

tswast: You're right.

I'd like to keep these instructions for folks that aren't using conda. I've moved the section after and added a disclaimer that it's not necessary if using conda.

docs/source/install.rst
```diff
@@ -32,6 +32,60 @@ Install from Source
...
+
+Conda
+~~~~~
+
+To install packages in development mode using conda, first
+install the conda build tools package in your root conda environment.
```
tswast: I've moved these instructions all to contributing.rst.

docs/source/install.rst
tswast: removed.

docs/source/install.rst
tswast: Done.

jreback:
yeah see #20 about the coverage issue. Maybe need to skip the user checks for this.

PR #83, opened 2017-08-21
ENH: Save BigQuery account credentials in a hidden user folder

### LINKED ISSUE #41: Provide a way to override the user account token location.
The user account token location is hard-coded to `bigquery_credentials.dat`. There should be a way to override this location.

parthea:
Closes #41

This PR adds ability for users to choose the location where BigQuery user account credentials will be stored via the `credentials_path` parameter.

This PR also changes the location of the file `bigquery_credentials.dat` that contains the BigQuery user account credentials. Previously this file was saved in the current working directory. The credentials file will now be stored in an application-specific hidden user folder on the operating system. On windows, this path is specified in the `APPDATA` environment variable as defined [here](https://technet.microsoft.com/en-ca/library/cc749104(v=ws.10).aspx) which is typically `'C:\Documents and Settings\username\Application Data\pandas-gbq\'` . Otherwise, credentials will be stored in `'~/.config/pandas-gbq/'`. If the file `bigquery_credentials.dat` is found in the current working directory, the file will be moved to the new location. 

@tswast @jreback Please could you take a look at your earliest convenience?

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-0.3.0 / 2017-??-??
+0.2.1 / 2017-??-??
 ------------------
 
 - :func:`read_gbq` now raises ``QueryTimeout`` if the request exceeds the ``query.timeoutMs`` value specified in the BigQuery configuration. (:issue:`76`)
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,12 +1,12 @@
 Changelog
 =========
 
-0.2.1 / 2017-??-??
+0.3.0 / 2017-??-??
 ------------------
 
 - :func:`read_gbq` now raises ``QueryTimeout`` if the request exceeds the ``query.timeoutMs`` value specified in the BigQuery configuration. (:issue:`76`)
 - Environment variable ``PANDAS_GBQ_CREDENTIALS_FILE`` can now be used to override the default location where the BigQuery user account credentials are stored. (:issue:`86`)
-
+- BigQuery user account credentials are now stored in an application-specific hidden user folder on the operating system. (:issue:`41`)
 
 0.2.0 / 2017-07-24
 ------------------
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -208,6 +208,7 @@ class GbqConnector(object):
         self.private_key = private_key
         self.auth_local_webserver = auth_local_webserver
         self.dialect = dialect
+        self.credentials_path = _get_credentials_file()
         self.credentials = self.get_credentials()
         self.service = self.get_service()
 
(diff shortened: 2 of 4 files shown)


pandas_gbq/gbq.py
```diff
@@ -300,14 +315,41 @@ def load_user_account_credentials(self):
...
+            config_path = os.environ['APPDATA']
+        else:
+            config_path = os.path.join(os.path.expanduser('~'), '.config')
+
+        config_path = os.path.join(config_path, 'pandas_gbq')
+
+        # Create a pandas_gbq directory in an application-specific hidden
+        # user folder on the operating system.
+        if not os.path.exists(config_path):
+            os.mkdir(config_path)
```
tswast: Is this recursive if .config doesn't exist?

pandas_gbq/gbq.py
```diff
@@ -300,14 +315,41 @@ def load_user_account_credentials(self):
...
+        """
+
+        import os
+
+        if os.name == 'nt':
+            config_path = os.environ['APPDATA']
+        else:
+            config_path = os.path.join(os.path.expanduser('~'), '.config')
+
+        config_path = os.path.join(config_path, 'pandas_gbq')
```
MaximilianR: You know this better than I do, but is there a case for having the default path be the default gcloud path, rather than anything specific to pandas?

pandas_gbq/gbq.py
parthea: Great catch! `os.mkdir` is not recursive. Fixed.

pandas_gbq/gbq.py
parthea: This is a great question! I've given this some thought and I don't feel comfortable writing default credentials into the `gcloud` path in case some users don't want to have their default `gcloud` credentials set (and I'm also worried about colliding with `gcloud`). Alternatively, users could run `gcloud auth application-default login` if they want to configure default credentials and the default `gcloud` credentials should be picked up and won't hit this code path. https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/gbq.py#L222

@tswast Thoughts?

parthea:
I've also made the changes required for pandas.io.gbq in the following branch:
https://github.com/pandas-dev/pandas/compare/master...parthea:sync-with-pandas-gbq

I'll hold off on creating a PR in pandas until this is merged.

pandas_gbq/gbq.py
tswast: +1 to not clobbering `gcloud` credentials. The token we save doesn't have the full Google Cloud Platform scope so that would break other applications that depend on those credentials.

If folks set up application default credentials it won't hit this code anyway.

MaximilianR: > users could run gcloud auth application-default login

> If folks set up application default credentials it won't hit this code anyway.

👍

parthea:
@tswast I've removed the `credentials_path` parameter in favour of the environment variable added in #87. I've rebased the changes with the latest master and everything appears to be working as expected in my local testing. Would you be able to take another look?

docs/source/changelog.rst
```diff
@@ -1,12 +1,12 @@
 Changelog
 =========
 
-0.2.1 / 2017-??-??
+0.3.0 / 2017-??-??
```
tswast: I wonder if we should do an incremental release before `0.3.0` if we plan to have that be the one where we swap to the google-cloud-python libraries? Seems like that PR has stalled and I don't have time to rebase it either.

parthea: Done

Issue #106, opened 2018-01-10
to_gbq result in UnicodeEncodeError


2legit:
Hi, I'm using Heroku to run a python based ETL process where I'm pushing the contents of a Pandas dataframe into Google BQ using to_gbq. However, it's generating a UnicodeEncodeError with the following stack trace, due to some non-latin characters. 

Strangely this works fine on my Mac but when I try to run it on Heroku, it's failing. It seems that for some reason, http.client.py is getting an un-encoded string rather than bytes and therefore, it's trying to encode with latin-1, which is the default but obviously would choke on anything non-latin, like Chinese chars.

2018-01-08T04:54:17.307496+00:00 app[run.2251]:
Load is 100.0% Complete044+00:00 app[run.2251]:
2018-01-08T04:54:20.443238+00:00 app[run.2251]: Traceback (most recent call last):
2018-01-08T04:54:20.443267+00:00 app[run.2251]:   File "AllCostAndRev.py", line 534, in <module>
2018-01-08T04:54:20.443708+00:00 app[run.2251]:     main(yaml.dump(data=ads_dict))
2018-01-08T04:54:20.443710+00:00 app[run.2251]:   File "AllCostAndRev.py", line 475, in main
2018-01-08T04:54:20.443915+00:00 app[run.2251]:     private_key=environ['skynet_bq_pk']
2018-01-08T04:54:20.443917+00:00 app[run.2251]:   File "/app/.heroku/python/lib/python3.6/site-packages/pandas_gbq/gbq.py", line 989, in to_gbq
2018-01-08T04:54:20.444390+00:00 app[run.2251]:     connector.load_data(dataframe, dataset_id, table_id, chunksize)
2018-01-08T04:54:20.444391+00:00 app[run.2251]:   File "/app/.heroku/python/lib/python3.6/site-packages/pandas_gbq/gbq.py", line 590, in load_data
2018-01-08T04:54:20.444653+00:00 app[run.2251]:     job_config=job_config).result()
2018-01-08T04:54:20.444656+00:00 app[run.2251]:   File "/app/.heroku/python/lib/python3.6/site-packages/google/cloud/bigquery/client.py", line 748, in load_table_from_file
2018-01-08T04:54:20.445248+00:00 app[run.2251]:     response = upload.transmit_next_chunk(transport)
[...]

PR #108, opened 2018-01-12
Encode before uploading

maxim-lian:
Potential fix for https://github.com/pydata/pandas-gbq/issues/106

...but someone with better bytes / str understanding needs to review

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -581,7 +581,11 @@ class GbqConnector(object):
                 self._print("\rLoad is {0}% Complete".format(
                     ((total_rows - remaining_rows) * 100) / total_rows))
 
-                body = BytesIO('{}\n'.format('\n'.join(rows)).encode('utf-8'))
+                body = '{}\n'.format('\n'.join(rows))
+                if isinstance(body, bytes):
+                    body = body.decode('utf-8')
+                body = body.encode('utf-8')
+                body = BytesIO(body)
 
                 try:
                     self.client.load_table_from_file(
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -581,7 +581,7 @@ class GbqConnector(object):
                 self._print("\rLoad is {0}% Complete".format(
                     ((total_rows - remaining_rows) * 100) / total_rows))
 
-                body = BytesIO('{}\n'.format('\n'.join(rows)).encode())
+                body = BytesIO('{}\n'.format('\n'.join(rows)).encode('utf-8'))
 
                 try:
                     self.client.load_table_from_file(
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -559,7 +559,7 @@ class GbqConnector(object):
 
     def load_data(self, dataframe, dataset_id, table_id, chunksize):
         from google.cloud.bigquery import LoadJobConfig
-        from six import StringIO
+        from six import BytesIO
 
         destination_table = self.client.dataset(dataset_id).table(table_id)
         job_config = LoadJobConfig()
@@ -581,7 +581,7 @@ class GbqConnector(object):
                 self._print("\rLoad is {0}% Complete".format(
                     ((total_rows - remaining_rows) * 100) / total_rows))
 
-                body = StringIO('{}\n'.format('\n'.join(rows)))
+                body = BytesIO('{}\n'.format('\n'.join(rows)).encode())
 
                 try:
                     self.client.load_table_from_file(
--- a/pandas_gbq/tests/test_gbq.py
+++ b/pandas_gbq/tests/test_gbq.py
@@ -1469,11 +1469,11 @@ class TestToGBQIntegrationWithLocalUserAccountAuth(object):
 
         assert result['num_rows'][0] == test_size
 
-
     def test_upload_unicode_data(self):
         test_id = "1"
         test_size = 10
-        df = DataFrame(np.random.randn(6, 4), index=range(6), columns=list('ABCD'))
+        df = DataFrame(np.random.randn(6, 4), index=range(6),
+                       columns=list('ABCD'))
         df.A = u'信用卡'
 
         gbq.to_gbq(df, self.destination_table + test_id, _get_project_id(),
@@ -1484,7 +1484,6 @@ class TestToGBQIntegrationWithLocalUserAccountAuth(object):
             project_id=_get_project_id())
 
         assert result['num_rows'][0] == test_size
-
 
 
 class TestToGBQIntegrationWithServiceAccountKeyContents(object):
--- a/pandas_gbq/tests/test_gbq.py
+++ b/pandas_gbq/tests/test_gbq.py
@@ -1157,8 +1157,8 @@ class TestToGBQIntegrationWithServiceAccountKeyPath(object):
                        _get_project_id(), private_key=_get_private_key_path())
 
     def test_upload_chinese_unicode_data(self):
-        test_id = "1"
-        test_size = 10
+        test_id = "2"
+        test_size = 6
         df = DataFrame(np.random.randn(6, 4), index=range(6),
                        columns=list('ABCD'))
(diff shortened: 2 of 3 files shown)


maxim-lian:
Travis passes. Not sure what's going on with codecov, but I think the PR should be good

pandas_gbq/tests/test_gbq.py
```diff
@@ -1154,6 +1156,43 @@ def test_google_upload_errors_should_raise_exception(self):
...
+        })
+
+        gbq.to_gbq(df, self.destination_table + test_id, _get_project_id(),
+                   chunksize=10000)
+
+        result = gbq.read_gbq("SELECT COUNT(*) AS num_rows FROM {0}".format(
+            self.destination_table + test_id),
+            project_id=_get_project_id())
+
+        assert result['num_rows'][0] == test_si
```
jasonqng: Typo: `test_size`

jasonqng:
Currently fails in my Python 2 env (error below) because the returned object from `'{}\n'.format('\n'.join(rows))` is a string, which thus fails to encode. Not sure if this is the most elegant way to handle this, but you can do a check on the joined object, and if it is a string, perform a decode to turn it into a unicode object:

```
                joined_rows = '{}\n'.format('\n'.join(rows))
                if isinstance(joined_rows, str):
                    joined_rows = joined_rows.decode('utf-8')
                body = BytesIO(joined_rows.encode('utf-8'))
```

Above tweak works for me, and both your tests now pass.

Error:
```
>>> from pandas import DataFrame
>>> import gbq
>>> df = DataFrame({
... 'string': ['Skywalker™', 'lego', 'hülle'],
... 'integer': [200, 300, 400],
... 'Date': [
...     '2017-12-13 17:40:39', '2017-12-13 17:40:39',
...     '2017-12-13 17:40:39'
... ]
... })
>>> gbq.to_gbq(df, "ad_doc.jn_test", project_id='XXXXX', if_exists='replace')



Load is 100% Complete
  File "gbq.py", line 989, in to_gbq
    connector.load_data(dataframe, dataset_id, table_id, chunksize)
  File "gbq.py", line 584, in load_data
    body = BytesIO('{}\n'.format('\n'.join(rows)).encode('utf-8'))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 63: ordinal not in range(128)
```

pandas_gbq/tests/test_gbq.py
jasonqng: Also, perhaps a test that downloads the uploaded dataframe and verifies that the special characters are preserved would be nice?

maxim-lian:
@jasonqng thank you v much. Even if a bit of a hack, I think it's good for the moment

pandas_gbq/tests/test_gbq.py
```diff
@@ -1154,6 +1156,45 @@ def test_google_upload_errors_should_raise_exception(self):
...
+
+        gbq.to_gbq(df, self.destination_table + test_id, _get_project_id(),
+                   chunksize=10000)
+
+        result = gbq.read_gbq("SELECT COUNT(*) AS num_rows FROM {0}".format(
+            self.destination_table + test_id),
+            project_id=_get_project_id())
+
+        assert result['num_rows'][0] == test_size
+        tm.assert_series_equal(result['string'], df['string'])
```
jasonqng: You'll want to either sort the `result` dataframe by integer or do an `order by` in the query to ensure that the assertion will pass (otherwise, the rows of result could be in a different order than your original `df`).

maxim-lian:
@jasonqng on second try, unfortunately that solution didn't work for Python3. I've pushed something that's not elegant either, but it works.

The one thing that doesn't work is the comparison when testing in Py2. If you look at the df it produces, it looks good, but comparing the strings is not successful. So I've skipped a subset of the test for moment

maxim-lian:
@tswast @jreback this is ready to go. Original reporter in https://github.com/pydata/pandas-gbq/issues/106 confirmed the fix works for them

pandas_gbq/gbq.py
```diff
@@ -581,7 +581,11 @@ def load_data(self, dataframe, dataset_id, table_id, chunksize):
                 self._print("\rLoad is {0}% Complete".format(
                     ((total_rows - remaining_rows) * 100) / total_rows))
 
-                body = StringIO('{}\n'.format('\n'.join(rows)))
+                body = '{}\n'.format('\n'.join(rows))
```
tswast: If you use `u'{}\n'.format()` is the if statement checking for `bytes` necessary?

tswast:
Please add this to the changelog at https://github.com/pydata/pandas-gbq/blob/master/docs/source/changelog.rst under the section heading `0.3.1 / (unreleased)`.

pandas_gbq/gbq.py
maxim-lian: Unfortunately not:
```

>               body = u'{}\n'.format(u'\n'.join(rows))
E               UnicodeDecodeError: 'ascii' codec can't decode byte 0xe4 in position 77: ordinal not in range(128)
```

I _think_ the nub problem is that `row.to_json` comes out as either bytes or str depending on the python version - so we need some branching somewhere. Unless there's a function in python that can deal with both (this all seems a bit inelegant)

(I also tried decoding the row first on 576, which made Py2 pass, but then python3 failed, because it can't decode unicode.

maxim-lian:
@tswast added whatsnew

tswast:
The build failed for this after merging because the test was waiting on user credentials:

```
pandas_gbq/tests/test_gbq.py::TestToGBQIntegrationWithServiceAccountKeyPath::test_upload_chinese_unicode_data Please visit this URL to authorize this application:
...
Enter the authorization code: 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

I'm not entirely sure why this is happening, since this test is in the `TestToGBQIntegrationWithServiceAccountKeyPath` class.

maxim-lian:
Hmmm. I did add some tests to that class. I've removed them in this branch: https://github.com/maxim-lian/pandas-gbq/tree/test-fix , (+ included a fix that I didn't carry over to those in the ServiceAccount class).

Could you try running that in Travis?

skion:
Confirming that this made the following (note the accented character) work again for me on Python 3.6.3... 
```python
pd.DataFrame({'my_string': ['María']}).to_gbq('xxx.yyy', 'project-id')
```

tswast:
@maxim-lian That looks like it should fix it, thanks. Build running at https://travis-ci.org/tswast/pandas-gbq/builds/330577200

Note: there are instructions at https://pandas-gbq.readthedocs.io/en/latest/contributing.html#running-google-bigquery-integration-tests for setting up your personal fork to build with integration tests on Travis.

maxim-lian:
Ah great! Not sure how I missed that. That's super

tswast:
Ugh. Still failing with a timeout on Travis.

```
pandas_gbq/tests/test_gbq.py::TestToGBQIntegrationWithServiceAccountKeyPath::test_upload_chinese_unicode_data 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

It's not asking for an authorization code anymore, so I'm not sure how it's getting stuck.

tswast:
I think I figured it out.

https://github.com/tswast/pandas-gbq/commit/fb6a2f6ac57d58f78cd1e2602860e73c8194af8e

These tests need a private key to be manually set. Building at https://travis-ci.org/tswast/pandas-gbq/builds/330598420

maxim-lian:
Ooof, sorry to leave you with that. https://github.com/pydata/pandas-gbq/issues/109 would solve all our problems forevermore

PR #111, opened 2018-01-19
TST: Use private key for unicode tests

tswast:
Fixes tests added in https://github.com/pydata/pandas-gbq/pull/108 for Travis.

/cc @maxim-lian


maxim-lian:
Thanks a lot @tswast

Issue #96, opened 2017-12-08
Do not manually loop over all rows when encoding a dataframe as JSON


tswast:
See: https://github.com/pydata/pandas-gbq/pull/25#discussion_r155355165

Currently the code loops over each row to encode it as JSON. This could be sped up by calling `to_json()` on the whole dataframe instead.

tswast:
Calling `to_json()` on the whole dataframe gives an actual JSON object, not newline-delimited JSON. This will become much easier if/when BigQuery supports a column-oriented data format such as Parquet.

tswast:
Cool, thanks @maxim-lian

tswast:
Ah, yeah CSV might be the way to go for now.

maxim-lian:
CSV also smaller! And no nesting possible so no loss there

But if json is preferred, I think you can use `orient=records`

tswast:
Looks like something like this should work to avoid having to write to a temporary file.

```python
def encode_chunk(df):
    """Return a file-like object of CSV-encoded rows.

    Args:
      df (pandas.DataFrame): A chunk of a dataframe to encode
    """
    csv_buffer = six.StringIO()
    df.to_csv(
        csv_buffer, index=False, header=False, encoding='utf-8',
        date_format='%Y-%m-%d %H:%M')

    # Convert to a BytesIO buffer so that unicode text is properly handled.
    # See: https://github.com/pydata/pandas-gbq/issues/106
    body = csv_buffer.getvalue()
    if isinstance(body, bytes):
        body = body.decode('utf-8')
    body = body.encode('utf-8')
    return six.BytesIO(body)
```

tswast:
Closed by https://github.com/pydata/pandas-gbq/pull/117

PR #117, opened 2018-02-10
BUG: Fix uploading of dataframes containing int64 and float64 columns

### LINKED ISSUE #116: BUG: Wrong JSON sent to BigQuery when both integer and float fields in schema
When both a float and integer field exist, integer fields (int64) are serialized as floats, resulting in an error from BigQuery (because in schema the field is defined as Integer).

```
df = pd.DataFrame([[1,1.1],[2,2.2] ],
                   index=['row 1', 'row 2'],
                   columns=['intColumn','floatColumn'])
 
# correct output
df.to_json()

# incorrect output (method used by pandas-gbq)
row = df.iloc[0]
row.to_json()

# correct output
df[['intColumn']].iloc[0].to_json()
```

tswast:
Fixes #116 and #96 by loading data in CSV chunks.

### DIFF
--- a/pandas_gbq/_load.py
+++ b/pandas_gbq/_load.py
@@ -1,0 +1,51 @@
+"""Helper methods for loading data into BigQuery"""
+
+from google.cloud import bigquery
+import six
+
+
+def encode_chunk(dataframe):
+    """Return a file-like object of CSV-encoded rows.
+
+    Args:
+      dataframe (pandas.DataFrame): A chunk of a dataframe to encode
+    """
+    csv_buffer = six.StringIO()
+    dataframe.to_csv(
+        csv_buffer, index=False, header=False, encoding='utf-8',
+        date_format='%Y-%m-%d %H:%M')
+
+    # Convert to a BytesIO buffer so that unicode text is properly handled.
+    # See: https://github.com/pydata/pandas-gbq/issues/106
+    body = csv_buffer.getvalue()
+    if isinstance(body, bytes):
+        body = body.decode('utf-8')
+    body = body.encode('utf-8')
+    return six.BytesIO(body)
+
+
+def encode_chunks(dataframe, chunksize):
+    dataframe = dataframe.reset_index(drop=True)
+    remaining_rows = len(dataframe)
+    total_rows = remaining_rows
+    start_index = 0
+    while start_index < total_rows:
+        chunk_buffer = encode_chunk(
+            dataframe[start_index:start_index+chunksize])
+        start_index += chunksize
+        remaining_rows = max(0, remaining_rows - chunksize)
+        yield remaining_rows, chunk_buffer
+
+
+def load_chunks(client, dataframe, dataset_id, table_id, chunksize):
+    destination_table = client.dataset(dataset_id).table(table_id)
+    job_config = bigquery.LoadJobConfig()
+    job_config.write_disposition = 'WRITE_APPEND'
+    job_config.source_format = 'CSV'
+
+    for remaining_rows, chunk_buffer in encode_chunks(dataframe, chunksize):
+        yield remaining_rows
+        client.load_table_from_file(
+            chunk_buffer,
+            destination_table,
+            job_config=job_config).result()
--- a/pandas_gbq/_load.py
+++ b/pandas_gbq/_load.py
@@ -24,26 +24,31 @@ def encode_chunk(dataframe):
     return six.BytesIO(body)
 
 
-def encode_chunks(dataframe, chunksize):
+def encode_chunks(dataframe, chunksize=None):
     dataframe = dataframe.reset_index(drop=True)
+    if chunksize is None:
+        yield 0, encode_chunk(dataframe)
+        return
+
     remaining_rows = len(dataframe)
     total_rows = remaining_rows
     start_index = 0
     while start_index < total_rows:
-        chunk_buffer = encode_chunk(
-            dataframe[start_index:start_index+chunksize])
+        end_index = start_index + chunksize
+        chunk_buffer = encode_chunk(dataframe[start_index:end_index])
         start_index += chunksize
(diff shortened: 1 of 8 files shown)


pandas_gbq/_load.py
```diff
@@ -0,0 +1,51 @@
...
+    # Convert to a BytesIO buffer so that unicode text is properly handled.
+    # See: https://github.com/pydata/pandas-gbq/issues/106
+    body = csv_buffer.getvalue()
+    if isinstance(body, bytes):
+        body = body.decode('utf-8')
+    body = body.encode('utf-8')
+    return six.BytesIO(body)
+
+
+def encode_chunks(dataframe, chunksize):
```
maxim-lian: Why the multiple chunks, rather than using a single chunk? Is it a memory issue? A UI / status bar updating issue?

pandas_gbq/tests/test__load.py
```diff
@@ -0,0 +1,26 @@
+
```
maxim-lian: FYI this file name currently has two underscores

pandas_gbq/_load.py
tswast: Because previously chunksize was required and I wasn't ready to make it option. I've just added a commit to this PR to make it optional. We'll want to update the default in Pandas after we release a package with this change.

pandas_gbq/tests/test__load.py
tswast: Thanks. I'm aware. I'm following the convention that the filename should be `test_` + the filename of the file under test.

tswast:
Full Travis build with system tests running at https://travis-ci.org/tswast/pandas-gbq/builds/339826702

tswast:
Ah, Travis uncovered a potential problem with using CSV.

```
=================================== FAILURES ===================================
 TestToGBQIntegrationWithServiceAccountKeyPath.test_upload_subset_columns_if_table_exists_append 
self = <pandas_gbq.tests.test_gbq.TestToGBQIntegrationWithServiceAccountKeyPath object at 0x7f28719af610>
    def test_upload_subset_columns_if_table_exists_append(self):
        # Issue 24: Upload is succesful if dataframe has columns
        # which are a subset of the current schema
        test_id = "16"
        test_size = 10
        df = make_mixed_dataframe_v2(test_size)
        df_subset_cols = df.iloc[:, :2]
    
        # Initialize table with sample data
        gbq.to_gbq(df, self.destination_table + test_id, _get_project_id(),
                   chunksize=10000, private_key=_get_private_key_path())
    
        # Test the if_exists parameter with value 'append'
        gbq.to_gbq(df_subset_cols,
                   self.destination_table + test_id, _get_project_id(),
>                  if_exists='append', private_key=_get_private_key_path())
pandas_gbq/tests/test_gbq.py:1096: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pandas_gbq/gbq.py:978: in to_gbq
    connector.load_data(dataframe, dataset_id, table_id, chunksize=chunksize)
pandas_gbq/gbq.py:572: in load_data
    self.process_http_error(ex)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ex = BadRequest(u'Error while reading data, error message: CSV table encountered too many errors, giving up. Rows: 1; errors: 1.',)
    @staticmethod
    def process_http_error(ex):
        # See `BigQuery Troubleshooting Errors
        # <https://cloud.google.com/bigquery/troubleshooting-errors>`__
    
>       raise GenericGBQException("Reason: {0}".format(ex))
E       GenericGBQException: Reason: 400 Error while reading data, error message: CSV table encountered too many errors, giving up. Rows: 1; errors: 1.
pandas_gbq/gbq.py:455: GenericGBQException
 TestToGBQIntegrationWithServiceAccountKeyPath.test_upload_data_flexible_column_order 
self = <pandas_gbq.tests.test_gbq.TestToGBQIntegrationWithServiceAccountKeyPath object at 0x7f2872fff310>
    def test_upload_data_flexible_column_order(self):
        test_id = "13"
        test_size = 10
        df = make_mixed_dataframe_v2(test_size)
    
        # Initialize table with sample data
        gbq.to_gbq(df, self.destination_table + test_id, _get_project_id(),
                   chunksize=10000, private_key=_get_private_key_path())
    
        df_columns_reversed = df[df.columns[::-1]]
    
        gbq.to_gbq(df_columns_reversed, self.destination_table + test_id,
                   _get_project_id(), if_exists='append',
>                  private_key=_get_private_key_path())
```

I think I probably need to include the schema definition in the load job, since we want to be able to upload a data frame even if the columns are out of order or there are extra columns.

maxim-lian:
> I think I probably need to include the schema definition in the load job, since we want to be able to upload a data frame even if the columns are out of order or there are extra columns.

Yes, for sure. I think you can do that fairly easily with `schema=_generate_bq_schema(df)`

(we have this but can't remember why we do the obj construction)

```python

def _bq_schema(df):
    schema_dict = _gbq._generate_bq_schema(df)
    schema = [bigquery.schema.SchemaField(x['name'], x['type'])
              for x in schema_dict['fields']]
    return schema
```

tswast:
I imagine you had to create the object construction as a workaround for https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4456

tswast:
Okay, I think I got it this time. Full build in-progress at https://travis-ci.org/tswast/pandas-gbq/builds/340579271

maxim-lian:
Congrats @tswast ! Thanks for pushing this through!

tswast:
Yeah! I'll plan to do a release this week to get all of these `to_gbq` fixes out there.

Issue #119, opened 2018-02-13
fail to read INT64 values on Windows


blose:
On pandas-gbq-0.3.0 it fails to get INT64 values:
`
new_df = gbq.read_gbq(
    "SELECT  12345678901234567 as iii",
    project_id=project_id,
    dialect="standard"
)
`

both on standard and legacy dialects. Querying from the real database does the same.

> Requesting query... ok.
> Job ID: 82d0024d-3e31-4518-a6d5-755cf96f108c
> Query running...
> Query done.
> Processed: 0.0 B Billed: 0.0 B
> Standard price: $0.00 USD
> 
> Retrieving results...
> Got 1 rows.
> 
> 
> ---------------------------------------------------------------------------
> OverflowError                             Traceback (most recent call last)
> <ipython-input-14-a22ed4823734> in <module>()
>       7 top10_active_users_df = gbq.read_gbq(
>       8     "SELECT  12345678901234567 as iii",
> ----> 9     project_id=project_id
>      10 )
> 
> ~\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\io\gbq.py in read_gbq(query, project_id, index_col, col_order, reauth, verbose, private_key, dialect, **kwargs)
>      98         private_key=private_key,
>      99         dialect=dialect,
> --> 100         **kwargs)
>     101 
>     102 
> 
> ~\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas_gbq\gbq.py in read_gbq(query, project_id, index_col, col_order, reauth, verbose, private_key, auth_local_webserver, dialect, **kwargs)
>     875                 field['mode'] != 'repeated':
>     876             final_df[field['name']] = \
> --> 877                 final_df[field['name']].astype(type_map[field['type'].upper()])
>     878 
>     879     connector.print_elapsed_seconds(
> 
> ~\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\util\_decorators.py in wrapper(*args, **kwargs)
>      89                 else:
>      90                     kwargs[new_arg_name] = new_arg_value
> ---> 91             return func(*args, **kwargs)
>      92         return wrapper
>      93     return _deprecate_kwarg
> 
> ~\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\core\generic.py in astype(self, dtype, copy, errors, **kwargs)
>    3408         # else, only a single dtype is given
[...]

PR #121, opened 2018-02-15
BUG: Explicitly use 64bit integers when converting from BQ types

blose:
Solved issue #119 
On 64-bit windows 10 the `int` values are cast into 32-bit integers for reasons given here
https://stackoverflow.com/questions/36278590/numpy-array-dtype-is-coming-as-int32-by-default-in-a-windows-10-64-bit-machine
This leads to unhandled exceptions when handling values beyond 32-bit signed range
In Bigquery, int is always 64bit, so the solution is to specify the data type explicitly


blose:
This is to solve the issue #119

maxim-lian:
I think that looks good! Thanks!

I don't think there's a helpful test we could add, given we don't test on Windows - @tswast ?

Issue #43, opened 2017-05-28
Support replacing partitions of date-partitioned tables


ghost:
This issue is about adding support to write to partitions of BigQuery's [date-partitioned tables](https://cloud.google.com/bigquery/docs/partitioned-tables). Currently, reading is supported but inserts fails during execution of `GbqConnector.delete_and_recreate`: partition is deleted but creation fails due to existing table.

PR would include some refactoring to the current `to_gbq` function (e.g. getting rid of duplicate calls to BQ API).

Please see [mremes/pandas-gbq](https://github.com/mremes/pandas-gbq) for current changes. Currently not ready for a PR as I've encountered one issue with the BigQuery API (with subsequent create, delete, insertAll calls, insertAll's changes are not seen but after long waiting time).

chalmerlowe:
This issue has seen no activity since 2017. Closing as will not fix.

PR #124, opened 2018-02-19
BUG: Add support to replace partitions in date-partitioned tables (#43)

mremes:
Relates to issue #43 

Changes:
* Ability to write into a partition of a date-partitioned table by the module inferring insert into a partitioned table from the table decorator or if specified by user
* Styling and docstring edits
* Fix use schema when appending data to a table with an existing schema (ie. regular appends or date-partitioned tables inserts)
Needs:
* Tests for:
  - stuff is in the right partition
  - only the target partition is affected
  - ?

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -337,10 +337,6 @@ class GbqConnector(object):
         This method authenticates using user credentials, either loading saved
         credentials from a file or by going through the OAuth flow.
 
-        Parameters
-        ----------
-        None
-
         Returns
         -------
         GoogleCredentials : credentials
@@ -567,7 +563,7 @@ class GbqConnector(object):
         try:
             for remaining_rows in _load.load_chunks(
                     self.client, dataframe, dataset_id, table_id,
-                    chunksize=chunksize):
+                    chunksize=chunksize, schema=schema):
                 self._print("\rLoad is {0}% Complete".format(
                     ((total_rows - remaining_rows) * 100) / total_rows))
         except self.http_error as ex:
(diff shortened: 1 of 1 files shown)


pandas_gbq/gbq.py
```diff
@@ -1028,24 +1030,30 @@ def exists(self, table_id):
...
+                return table.num_rows > 0
+
             return True
         except NotFound:
             return False
         except self.http_error as ex:
             self.process_http_error(ex)
 
-    def create(self, table_id, schema):
+    def create(self, table_id, schema, date_partitioned=False):
```
tswast: Rather than pass in `date_partitioned` here, I'd like to see an argument similar to the configuration argument in `read_gbq` which can take a dictionary in the JSON API format. That way options like `timePartitioning` can be set.

pandas_gbq/gbq.py
```diff
@@ -699,8 +695,9 @@ def delete_and_recreate_table(self, dataset_id, table_id, table_schema):
         table = _Table(self.project_id, dataset_id,
                        private_key=self.private_key)
         table.delete(table_id)
```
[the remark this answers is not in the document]
mremes: I tried it with the decorated table name and it worked as intended.

pandas_gbq/gbq.py
mremes: That makes sense to me! I'll look into including the configuration in the `to_gbq` call to enable this and rest of the API configuration.

tswast:
Closing as pandas-gbq has changed quite a bit since this PR was opened.

PR #125, opened 2018-02-20
Test incorrectly named

maxim-lian:
I don't think this was running - it wasn't when I ran `pytest` locally


tswast:
Good catch, thanks! For as much time as I've spent in `test_gbq.py`, I never knew we had unit tests in that file.

Since this will leave the Travis CI tests busted for a while, I've created a new `unit-tests` branch for us to fix the rest of them before merging this into `master`.

PR #126, opened 2018-02-20
fix test_parse test

maxim-lian:
I don't think this test was passing for a while?


maxim-lian:
This leaves other broken tests from https://github.com/pydata/pandas-gbq/pull/125, but at least fixes one

Issue #72, opened 2017-07-11
Integration tests should fail rather than skip if project id is not set in Travis


parthea:
Certain integration tests are [skipped](https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/tests/test_gbq.py#L25) if a BigQuery project is not set in Travis. My initial thought is that the tests should fail if a BigQuery project id is not provided. 

For reference, here are the steps to run the BigQuery integration tests on Travis:
https://pandas-gbq.readthedocs.io/en/latest/contributing.html#running-google-bigquery-integration-tests

max-sixty:
This was merged to master earlier this year, closing

PR #127, opened 2018-02-20
project_id as optional

maxim-lian:
The underlying Google libraries can read default projects, so there's no need for us to force users to supply one


pandas_gbq/gbq.py
```diff
@@ -761,7 +761,7 @@ def read_gbq(query, project_id=None, index_col=None, col_order=None,
     ----------
     query : str
         SQL-Like Query to return data values
-    project_id : str
+    project_id : str (optional)
```
tswast: The `project_id` is still required when using user authentication, so we should clarify that it is only optional if using default credentials.

Or maybe also optional with a service account? I think we might have to parse as JSON and grab the project ID from the service account and use https://google-auth.readthedocs.io/en/latest/reference/google.oauth2.service_account.html#google.oauth2.service_account.Credentials.from_service_account_info for that to work.

maxim-lian: Thanks for bearing with me on the authentication questions

> The project_id is still required when using user authentication

Is this right? If I'm running localling with user auth, `google.auth` can pick up my default project fine:

```
In [20]: import google.cloud.bigquery

In [21]: c=google.cloud.bigquery.Client()

In [22]: c
Out[22]: <google.cloud.bigquery.client.Client at 0x1138e44a8>

In [23]: c.project
Out[23]: 'sixty-capital'
```

Auth:
```
gcloud auth list                                             
       Credentialed Accounts
ACTIVE  ACCOUNT
*       []@sixtycapital.com

To set the active account, run:
    $ gcloud config set account `ACCOUNT`
```

tswast: `google.cloud.bigquery.Client()` does user auth using the Cloud SDK. I'm referring to when the Cloud SDK is not installed, where pandas-gbq does it's user auth flow.

tswast: To be more specific the credential creation at

https://github.com/pydata/pandas-gbq/blob/16747609dc58f0b6df2c4e896db5a3f365c4677c/pandas_gbq/gbq.py#L272-L283

does not include a default project.

maxim-lian: Bear with me once more:
Why don't we add it? It's in the JSON file:

image

tswast: In 3LO there are two projects we are concerned with:

- The client secrets project. In this case, it is the project for Pandas GBQ. It has the BigQuery API enabled, but users shouldn't have permissions to run queries against this project. If we did use it, it would bill the Pandas team, not the user.
- The user's project. We know this in the case of a service account or when using the `gcloud` command, but we don't know it if using client secrets.

maxim-lian: OK I see now. Because we do the `app_flow`, those creds don't come with a project.
Though not sure what 3LO is

tswast: 3LO - 3-legged OAuth 2. (The protocol that `app_flow` follows.)

pandas_gbq/tests/test_gbq.py
```diff
@@ -87,9 +78,12 @@ def _test_imports():
...
 
 
-@pytest.fixture
-def project():
-    return _get_project_id()
+@pytest.fixture(params=['env'])
+def project(request):
+    if request.param == 'env':
+        return _get_project_id()
+    elif request.param == 'none':
```
maxim-lian: (this didn't get used, but could param over it in the future)

maxim-lian:
Updated!

pandas_gbq/gbq.py
```diff
@@ -7,6 +7,7 @@
 from distutils.version import StrictVersion
 from time import sleep
 
+import google
```
tswast: `google` is a namespace package for many different packages. Could you import `google.auth` instead?

tswast: Also, we don't tend to import our direct dependencies here, since we want the code for imports checking to run before crashing on a failed import.

pandas_gbq/gbq.py
```diff
@@ -170,6 +170,9 @@ def __init__(self, project_id, reauth=False,
         from google.api_core.exceptions import GoogleAPIError
         from google.api_core.exceptions import ClientError
         self.http_error = (ClientError, GoogleAPIError)
+        if not project_id:
+            from google.auth import default
+            _, project_id = default()
```
tswast: On second thought, this could cause us to go through the whole "application default credentials" flow twice, which could be quite slow. (It checks for env var, checks for GCE, checks for App Engine, ...)

tswast: Could we change https://github.com/maxim-lian/pandas-gbq/blob/a7f6c43434bc21e1d2fc7633d3c5cb9c04060a4a/pandas_gbq/gbq.py#L223 to not discard the project instead?

maxim-lian: Yeah not great:
```

In [4]: %timeit google.auth.default()
573 ms ± 14.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

maxim-lian: What do you think about changing those methods to return 2-tuples?

tswast: I think tuples would be reasonable. The other (more disruptive, probably) option would be to pass in the project to the credentials methods and have them return a `google.cloud.bigquery.Client`.

maxim-lian:
This will close https://github.com/pydata/pandas-gbq/issues/72

pandas_gbq/gbq.py
tswast: I updated this PR to tuples as we discussed.

maxim-lian:
Looks like you're updating @tswast - thanks. I shouldn't have let this hang around for so long - then I hit merge conflicts and slowed down even more

Feel free to discard anything ancillary to the main task here (e.g. specifying the auth type in the test run rather than the hacky function that skip in travis), and I'll add them separately

Thanks!

tswast:
Cool. I'll get this to the finish line. Thanks for your work so far, @maxim-lian 

Re: auth type tests. Yeah, I'll revert back to the wonky `_skip_in_travis()` functions for now. I'm thinking of doing a refactor soon that pull the auth functions out of `GbqConnector` and into their own module. Then I think it'll make more sense to test auth separately from read/to.

Issue #123, opened 2018-02-17
ValueError: setting an array element with a sequence.


MtDersvan:
INFO:
`pandas_gbq.__version__ : '0.3.0', '0.3.1'`
`python`:  `3.6.2,  3.5.2`

SNIPPET:
~~~python
from pandas.io import gbq
df  = gbq.read_gbq(
    """
    #standardSQL
    SELECT embedding_v1
    FROM `{TABLE_ID}` LIMIT 10
    """.format(TABLE_ID='patents-public-data.google_patents_research.publications'),
    dialect='standard',
    project_id='XXXXXXX',
    configuration={'query': {'useQueryCache': True}}
    )
~~~
ERROR:
~~~console
Traceback (most recent call last):
  File "<stdin>", line 9, in <module>
  File "/Users/xxxxxx/tf/lib/python3.6/site-packages/pandas/io/gbq.py", line 99, in read_gbq
    **kwargs)
  File "/Users/xxxxxx/tf/lib/python3.6/site-packages/pandas_gbq/gbq.py", line 828, in read_gbq
    final_df = _parse_data(schema, rows)
  File "/Users/xxxxxx/tf/lib/python3.6/site-packages/pandas_gbq/gbq.py", line 729, in _parse_data
    page_array[row_num][col_num] = field_value
ValueError: setting an array element with a sequence.
~~~
ISSUE:
`'patents-public-data.google_patents_research.publications'` - a public dataset.
`'embedding_v1'` - a repeated float field.
Google BigQuery tools parses this query without a problem, but pandas-gbq outputs the upper-mentioned issue. 
Maybe related to https://github.com/pydata/pandas-gbq/pull/101

tswast:
Interesting. Maybe a regression from #101? Or maybe those tests were insufficient?

jasonqng:
Confirmed it indeed fails and only on arrays with floats. For example:
```
gbq.read_gbq("select [1.1,2.2,3.3]","project",dialect="standard")
```

Works for all other types except for timestamps, for which it fails as well:
```
gbq.read_gbq("select [TIMESTAMP_SECONDS(1),TIMESTAMP_SECONDS(2),TIMESTAMP_SECONDS(3)]","creator-insights",dialect="standard")
```

Leads me to suspect it's this custom dtype handling we're doing here in [_parse_data](https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/gbq.py#L711):
```
[...]

PR #128, opened 2018-02-20
Faster dataframe construction

max-sixty:
This is an attempt to push the DataFrame construction down to Pandas, which is extremely fast, rather than looping over the entire dataframe in python ourselves

One downside is that the DataFrame doesn't accept multiple dtypes, so the resulting dataframe will be defined only by the data.

We could run `.astype` on each column, which I think would be fairly performant. But we'd need to write around cases such as non-float/obj columns having nulls. And the DataFrame constructor likely gets it right a vast majority of the time


maxim-lian:
> We could run .astype on each column, which I think would be fairly performant. But we'd need to write around cases such as non-float/obj columns having nulls. And the DataFrame constructor likely gets it right a vast majority of the time

I added this.

Though needs a review, as it's possible I haven't thought of a corner case

pandas_gbq/gbq.py
```diff
@@ -716,19 +715,16 @@ def _parse_data(schema, rows):
...
-    for row_num, entries in enumerate(rows):
-        for col_num in range(len(col_types)):
-            field_value = entries[col_num]
-            page_array[row_num][col_num] = field_value
-
-    return DataFrame(page_array, columns=col_names)
+
+    column_dtypes = {
+        str(field['name']):
+        dtype_map.get(field['type'].upper(), object) for field in fields
```
tswast: I wonder if our `dtype_map` needs some updating, as `google-cloud-bigquery` already does some type conversions. Travis seems to think something is up with this.

```
pandas_gbq/gbq.py:853: in read_gbq
    final_df[field['name']].astype(type_map[field['type'].upper()])
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/pandas/core/generic.py:3054: in astype
    raise_on_error=raise_on_error, **kwargs)
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/pandas/core/internals.py:3189: in astype
    return self.apply('astype', dtype=dtype, **kwargs)
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/pandas/core/internals.py:3056: in apply
    applied = getattr(b, f)(**kwargs)
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/pandas/core/internals.py:461: in astype
    values=values, **kwargs)
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/pandas/core/internals.py:504: in _astype
    values = _astype_nansafe(values.ravel(), dtype, copy=True)
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/pandas/types/cast.py:534: in _astype_nansafe
    return lib.astype_intsafe(arr.ravel(), dtype).reshape(arr.shape)
pandas/lib.pyx:980: in pandas.lib.astype_intsafe (pandas/lib.c:17409)
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   ???
E   ValueError: invalid literal for long() with base 10: '2017-12-13 17:40:39'
```

Sidenote (note relevant for this PR): I think this line may be important for https://github.com/pydata/pandas-gbq/issues/123. We'd want to fall back to object if the mode is `REPEATED`.

max-sixty:
This now works - was actually very close to working prior.

This is probably 40% of the way to the 'load in each page into an array and then concat at the end'. 20% of the work though; so a decent place to pause & merge

pandas_gbq/gbq.py
```diff
@@ -444,29 +443,30 @@ def _get_credentials_file():
...
+            yield name, dtype
+
+
+def _parse_data(schema, rows):
+
+    column_dtypes = OrderedDict(_parse_schema(schema['fields']))
+
+    df = DataFrame(data=(iter(r) for r in rows), columns=column_dtypes.keys())
+    for column in df:
+        df[column] = df[column].astype(column_dtypes[column])
```
tswast: Do we even need the `astype`? I recall that when I dropped it last time I encountered https://github.com/pydata/pandas-gbq/issues/174. A behavior change like that would require a version bump, though, so I'm happy with keeping this a 0.6.1 release if you'd prefer to get this out as-is.

max-sixty: That's a good point re `int` with nulls. I'll add a test and see what happens.

I think we do need this for dates; otherwise they'll be `str`

max-sixty: Actually there's already a test: https://github.com/max-sixty/pandas-gbq/blob/no-python-loop/tests/system/test_gbq.py#L165

What's happening: pandas is ignoring setting a column with `NaN` to `int`, and keeping it as object

PR #131, opened 2018-02-20
Fix use schema from params

mremes:
Issue #132


maxim-lian:
Can we add a test?

maxim-lian:
@mremes thank you for finding the error & fixing. #150 added tests too, so closing this

PR #134, opened 2018-02-21
Fix array of floats bug

jasonqng:
This PR resolves the bug documented here: https://github.com/pydata/pandas-gbq/issues/123

Basically, the custom dtype handling we have to convert missing values to NaNs fails when the field in BigQuery is an array (that is, in repeated mode). We simply set the dtype to an object if it is an array.

Added a test to check for this in the future. All previous tests still pass. Can also verify the bug and that this fixes by running:

```
gbq.read_gbq("select [1.1,2.2,3.3]","project",dialect="standard")
```

@tswast @MtDersvan

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+0.3.2 / [TBD]
+------------------
+- Fix bug with querying for an array of floats (:issue:`123`)
+
 0.3.1 / 2018-02-13
 ------------------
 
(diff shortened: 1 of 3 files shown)


tswast:
Oh, and mind adding an entry to the [changelog](https://github.com/pydata/pandas-gbq/blob/master/docs/source/changelog.rst)? This would be for version `0.3.2 / [TBD]`

jasonqng:
@tswast Hm, most of the tests were skipped, was there something I was supposed to do? https://travis-ci.org/jasonqng/pandas-gbq/builds/345048014 

Pytest runs on my local and everything passes here.

jasonqng:
@tswast All tests pass: https://travis-ci.org/jasonqng/pandas-gbq/builds/345048014

PR #139, opened 2018-03-03
Better test structure

maxim-lian:
Move towards more pytest idioms, reduce amount of code, reduce branching. This is a starter; still lots to do.

But can merge if ppl agree and keep making small improvments

### DIFF
--- a/pandas_gbq/tests/test_gbq.py
+++ b/pandas_gbq/tests/test_gbq.py
@@ -378,46 +378,39 @@ def test_should_read(project, credentials):
     tm.assert_frame_equal(df, DataFrame({'valid_string': ['PI']}))
 
 
-class TestReadGBQIntegrationWithServiceAccountKeyPath(object):
-
-    @classmethod
-    def setup_class(cls):
-        # - GLOBAL CLASS FIXTURES -
-        #   put here any instruction you want to execute only *ONCE* *BEFORE*
-        #   executing *ALL* tests described below.
-
-        _skip_if_no_private_key_path()
+class TestReadGBQIntegration(object):
 
     @pytest.fixture(autouse=True)
-    def setup(self, project):
+    def setup(self, project, credentials):
         # - PER-TEST FIXTURES -
         # put here any instruction you want to be run *BEFORE* *EVERY* test is
         # executed.
         self.gbq_connector = gbq.GbqConnector(
-            project, private_key=_get_private_key_path())
+            project, private_key=credentials)
+        self.credentials = credentials
 
     def test_should_properly_handle_valid_strings(self):
         query = 'SELECT "PI" AS valid_string'
         df = gbq.read_gbq(query, project_id=_get_project_id(),
-                          private_key=_get_private_key_path())
+                          private_key=self.credentials)
         tm.assert_frame_equal(df, DataFrame({'valid_string': ['PI']}))
 
     def test_should_properly_handle_empty_strings(self):
         query = 'SELECT "" AS empty_string'
         df = gbq.read_gbq(query, project_id=_get_project_id(),
-                          private_key=_get_private_key_path())
+                          private_key=self.credentials)
         tm.assert_frame_equal(df, DataFrame({'empty_string': [""]}))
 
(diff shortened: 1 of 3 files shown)


maxim-lian:
More improvements made, at a good place to merge now

maxim-lian:
@tswast @parthea Could you review? Let's get these in and then I'll merge them into the others.

pandas_gbq/tests/test_gbq.py
```diff
@@ -179,39 +172,80 @@ def test_generate_bq_schema_deprecated():
         gbq.generate_bq_schema(df)
 
 
-class TestGBQConnectorIntegrationWithLocalUserAccountAuth(object):
+@pytest.fixture(params=['local', 'service_path', 'service_creds'])
```
tswast: Woah. TIL about [parameterized fixtures](https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures). Cool stuff.

pandas_gbq/tests/test_gbq.py
```diff
@@ -402,129 +371,61 @@ def test_read_gbq_with_corrupted_private_key_json_should_fail(self):
...
-
-        _skip_if_no_project_id()
-        _skip_if_no_private_key_path()
 
-        _setup_common()
+class TestReadGBQIntegration(object):
 
-    def setup_method(self, method):
+    @pytest.fixture(autouse=True)
+    def setup(self, project, credentials):
```
tswast: Since queries can be pretty long-running, I don't think we really want to use all 3+ credentials types here.

pandas_gbq/tests/test_gbq.py
```diff
@@ -402,129 +371,61 @@ def test_read_gbq_with_corrupted_private_key_json_should_fail(self):
...
-
-    def teardown_method(self, method):
-        # - PER-TEST FIXTURES -
-        # put here any instructions you want to be run *AFTER* *EVERY* test is
-        # executed.
-        pass
-
-    def test_should_read_as_user_account(self):
-        _skip_local_auth_if_in_travis_env()
+def test_should_read(project, credentials):
```
tswast: As opposed to my comment below, I do like this test using all three credentials types.

I think it's good to verify that with a speedy query like this one.

pandas_gbq/tests/test_gbq.py
```diff
@@ -968,72 +871,52 @@ def test_array_agg(self):
...
-        # executing *ALL* tests described below.
-
-        _skip_if_no_project_id()
-        _skip_if_no_private_key_path()
-
-        _setup_common()
-
-    def setup_method(self, method):
+    @pytest.fixture(autouse=True, scope='function')
+    def setup(self, project, credentials):
```
tswast: Ditto. I don't think we want all tests running with all credentials types, but we should have 1 test that does.

maxim-lian:
> I have some worries this will make the tests take longer 

Generally only one of the sets of tests will run, because for the other auth types they'll be skipped. 
https://github.com/pydata/pandas-gbq/pull/139/files#diff-a348bd9d754cdc9d2ed5ec75a9626bcdR186

Indeed - this is no worse than the existing setup, where we had a bunch of copy & pasted code, but only one of the versions would run because the other would be skipped. Here we can also run tests using user auth, but that's the only growth in available tests.

In a future version, we could pass explicit options from the command line on which auth type(s) should be used, rather than auto-skipping based on the environment. That would also have the advantage that the test suite won't pass if no authentication is available.

But I don't think we need to couple those changes - this is strictly an improvement from the existing version - or is there something I'm missing?

tswast:
@maxim-lian Did this pass on your Travis?

maxim-lian:
> so in that case both will run using service account keys.

That's a good point. Should we skip one in Travis?

Checking [here](https://travis-ci.org/maxim-lian/pandas-gbq/builds/348849353), they do take about twice as long

They do pass

maxim-lian:
I've skipped one in Travis. Would be better to make the travis run command pass a command line which defined this to a parameterized test suite, but this is fine for the moment IMO

Issue #53, opened 2017-06-07
TST: Potentially add a unit test to verify the functionality of parameter `auth_local_webserver`


parthea:
It may be possible to add a unit test to verify the functionality of parameter `auth_local_webserver` of `read_gbq()` and `to_gbq()` in #39 in order to improve testing. One potential solution is to capture the output in the unit test and look for specific text :

`'Go to the following link in your browser:'` , when `auth_local_webserver=False`
or
`'Your browser has been opened to visit:'` , when `auth_local_webserver=True`

tswast:
This parameter is now passed on through to pydata-google-auth. I've filed this as https://github.com/pydata/pydata-google-auth/issues/9 over there.

PR #140, opened 2018-03-03
Remove console flow

maxim-lian:
The Console Flow is a strict subset of the functionality of the Local Webserver - if the browser fails to open, it's exactly the same. It makes writing parameterized tests harder, and is a small contributor to overall gunk

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -4,6 +4,9 @@ Changelog
 0.3.2 / [TBD]
 ------------------
 - Fix bug with querying for an array of floats (:issue:`123`)
+- Removed Console Flow from user auth, making Local Webserver the default.
+  When attempting to auth, a browser will attempted to be opened. Should that
+  fail, you can copy and paste the URL as before
 
 0.3.1 / 2018-02-13
 ------------------
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -174,7 +174,7 @@ class GbqConnector(object):
         self.project_id = project_id
         self.reauth = reauth
         self.verbose = verbose
-        self.private_key = private_key©
+        self.private_key = private_key
         self.dialect = dialect
         self.credentials_path = _get_credentials_file()
         self.credentials = self.get_credentials()
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -1,15 +1,14 @@
+import json
+import os
+import sys
+import time
 import warnings
 from datetime import datetime
-import json
-import time
+from distutils.version import StrictVersion
 from time import sleep
-import sys
-import os
 
 import numpy as np
-
-from distutils.version import StrictVersion
-from pandas import compat, DataFrame
+from pandas import DataFrame, compat
 from pandas.compat import lzip
 
 
@@ -168,16 +167,14 @@ class GbqConnector(object):
     scope = 'https://www.googleapis.com/auth/bigquery'
 
     def __init__(self, project_id, reauth=False, verbose=False,
-                 private_key=None, auth_local_webserver=False,
-                 dialect='legacy'):
+                 private_key=None, dialect='legacy'):
         from google.api_core.exceptions import GoogleAPIError
         from google.api_core.exceptions import ClientError
         self.http_error = (ClientError, GoogleAPIError)
         self.project_id = project_id
         self.reauth = reauth
         self.verbose = verbose
-        self.private_key = private_key
-        self.auth_local_webserver = auth_local_webserver
+        self.private_key = private_key©
         self.dialect = dialect
         self.credentials_path = _get_credentials_file()
         self.credentials = self.get_credentials()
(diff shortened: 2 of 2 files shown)


maxim-lian:
Would close https://github.com/pydata/pandas-gbq/issues/53

parthea:
Did you have any issue getting this to work on a headless machine? I cloned this branch on my headless machine and received an exception that the web server was already in use when trying to authenticate.  I like that the console flow allows you to easily authenticate in terminal if you're using `pandas-gbq` on a headless machine. 

```
>>> gbq.read_gbq('select 1','12345')
  File "pandas_gbq/gbq.py", line 818, in read_gbq
    dialect=dialect)
  File "pandas_gbq/gbq.py", line 180, in __init__
    self.credentials = self.get_credentials()
  File "pandas_gbq/gbq.py", line 194, in get_credentials
    credentials = self.get_user_account_credentials()
  File "pandas_gbq/gbq.py", line 367, in get_user_account_credentials
    credentials = app_flow.run_local_server()
  File "/usr/local/lib/python2.7/dist-packages/google_auth_oauthlib/flow.py", line 407, in run_local_server
    host, port, wsgi_app, handler_class=_WSGIRequestHandler)
  File "/usr/lib/python2.7/wsgiref/simple_server.py", line 151, in make_server
    server = server_class((host, port), handler_class)
  File "/usr/lib/python2.7/SocketServer.py", line 417, in __init__
    self.server_bind()
  File "/usr/lib/python2.7/wsgiref/simple_server.py", line 48, in server_bind
    HTTPServer.server_bind(self)
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 108, in server_bind
    SocketServer.TCPServer.server_bind(self)
  File "/usr/lib/python2.7/SocketServer.py", line 431, in server_bind
    self.socket.bind(self.server_address)
  File "/usr/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
socket.error: [Errno 98] Address already in use
```

maxim-lian:
That's v weird. It works fine for me in headless:

```
In [1]: import pandas.io.gbq as gbq

In [2]: gbq.read_gbq('select 1 as a', project_id='sixty-capital', auth_local_webserver=True)
Please visit this URL to authorize this application: https://accounts.google.com/o/oauth2/auth?response_type=code&client_id=495642085510-k0tmvj2m941jhre2nbqka17vqpjfddtd.apps.googleusercontent.com&redirect_uri=http%3A%2F%2Flocalhost%3A8080%2F&scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fbigquery&state=b0aQA29ZDoyHys[]PIViG&access_type=offline
```

Is yours a one-off or something repeatable? If the latter, we could deprecate and change the default, but leave the `False` option and see if it's widespread?

tswast:
@craigcitro I believe I put the console flow in based on a conversation with you. Thoughts on this change?

craigcitro:
In general, assuming that the user is at the same machine as the one running their python code is a faulty assumption IMO. That is, I don't believe that the local webserver flow handles all cases that the console flow does.

In particular, here's a simple recipe to reproduce: start a VM anywhere with only SSH access, connect from your laptop via SSH, and run `pandas-gbq`-using code there. Assuming you have not or cannot set up SSH forwarding of the right ports, there's no way for your laptop's web browser to talk to the local webserver running on your VM.

Or, a repro closer to my heart: open https://colab.research.google.com and try from there. There's no way for your local browser to talk to a random webserver started on that machine. (This is one of the reasons that we have some custom auth code in play.)

maxim-lian:
@craigcitro when I try in a docker container (https://github.com/pydata/pandas-gbq/pull/140#issuecomment-370248513), it prints out a web address whether or not I pass `True` or `False` to `auth_local_webserver`, because it can't find a browser.

If it doesn't have that in some situations, then I 100% agree with you. My point is based only there being a fallback, rather than whether that fallback is required.

craigcitro:
@maxim-lian It will always print out a web address; one of the query parameters in that address is the `redirect_uri`.

* If that uri is a real URL, eg `http://localhost:12345/...`, then when you click "authorize", your browser will send back a response containing your new oauth2 token to that URL. If that web request fails, there's no way to complete the authorization dance.

* If that uri is the "oauth2 OOB uri" (`urn:ietf:wg:oauth:2.0:oob`), then when you click "authorize", you'll get a code you can copy-paste back to the calling application. This application can then make a request to the oauth2 server (`accounts.google.com` in this case) to exchange that code for the oauth2 token and related bits.

So if you opt to go with the local server flow, and there's no way for your browser to communicate with that "local" server, then you can no longer do user-based authentication.

maxim-lian:
@craigcitro Ah! I see. Even though the initial link is to `accounts.google.com`, it needs to redirect to `localhost`. Thanks for bearing with me.

Would you prefer the default stayed as-is?

craigcitro:
@maxim-lian No problem at all! 😁 I've been doing oauth2-related things since, like 2011, and I still get confused regularly. (But hey, job security!)

I think *not* using a local webserver is the right default. My $0.02:
* local webserver: smoother when it works, but utterly confusing when it doesn't
* oob flow: clunkier, but always works

I've copy-pasted enough auth codes to last me a lifetime, but I don't fancy users getting stuck. WDYT?

maxim-lian:
Super, agree. 

Closing

Issue #73, opened 2017-07-14
read_gbq() unnecessarily waiting on getting default credentials from Google


dfontenot:
When attempting to grant pandas access to my GBQ project, I am running into an issue where `read_gbq` is trying to get default credentials, failing / timing out, then printing out a URL to go to to grant the credentials. Since I'm not running this on google cloud platform, I do not expect to be able to get default credentials. In my case, I only want to run the CLI flow (without having oauth call back to my local server).

Here's the code
```
>>> import pandas_gbq as gbq
>>> gbq.read_gbq('SELECT 1', project_id=<project_id>, auth_local_webserver=False)
```

Here's what I see when I trigger a SIGINT once the query is invoked:
```
  File "/usr/lib/python3.5/site-packages/pandas_gbq/gbq.py", line 214, in get_credentials
    credentials = self.get_application_default_credentials()
  File "/usr/lib/python3.5/site-packages/pandas_gbq/gbq.py", line 243, in get_application_default_credentials
    credentials, _ = google.auth.default(scopes=[self.scope])
  File "/usr/lib/python3.5/site-packages/google/auth/_default.py", line 277, in default
    credentials, project_id = checker()
  File "/usr/lib/python3.5/site-packages/google/auth/_default.py", line 274, in <lambda>
    lambda: _get_gce_credentials(request))
  File "/usr/lib/python3.5/site-packages/google/auth/_default.py", line 176, in _get_gce_credentials
    if _metadata.ping(request=request):
  File "/usr/lib/python3.5/site-packages/google/auth/compute_engine/_metadata.py", line 73, in ping
    timeout=timeout)
  File "/usr/lib/python3.5/site-packages/google/auth/transport/_http_client.py", line 103, in __call__
    method, path, body=body, headers=headers, **kwargs)
  File "/usr/lib/python3.5/http/client.py", line 1106, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python3.5/http/client.py", line 1151, in _send_request
    self.endheaders(body)
  File "/usr/lib/python3.5/http/client.py", line 1102, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python3.5/http/client.py", line 934, in _send_output
    self.send(msg)
  File "/usr/lib/python3.5/http/client.py", line 877, in send
    self.connect()
  File "/usr/lib/python3.5/http/client.py", line 849, in connect
    (self.host,self.port), self.timeout, self.source_address)
[...]

Issue #136, opened 2018-02-28
Hide google auth warning


melissachang:
Every time I use pandas_gbq, a google auth warning is printed. The warning is harmless; pandas_gbq still works. But it would be nice if the warning wasn't printed. Perhaps pandas_gbq could set GOOGLE_CLOUD_PROJECT.

```python
import logging

import pandas as pd
import pandas_gbq


# Log to stderr.
logging.basicConfig(level=logging.WARNING,
    format='%(asctime)s %(filename)10s:%(lineno)s %(levelname)s %(message)s',
    datefmt='%Y%m%d%H:%M:%S')
logger = logging.getLogger('warning')


def main():
  project_id = 'MY_PROJECT'
  dataset_id = 'MY_DATASET'
  table_name = 'MY_TABLE'
  df = pd.read_gbq(
      'SELECT * FROM `%s.%s.%s`' % (project_id, dataset_id, table_name),
      project_id=project_id, dialect='standard')
  print('%d rows read.' % len(df))
  print('pandas version %s' % pandas_gbq.__version__)

if __name__ == '__main__':
  main()
```

```
2018022815:32:50 _default.py:280 WARNING No project ID could be determined. Consider running `gcloud config set project` or setting the GOOGLE_CLOUD_PROJECT environment variable
Requesting query... ok.
Job ID: 82276626-c62f-4e58-ba0b-2adba047c7bd
Query running...
Query done.
Cache hit.

Retrieving results...
Got 901 rows.

Total time taken 0.91 s.
Finished at 2018-02-28 15:32:52.
901 rows read.
pandas version 0.3.1
```

maxim-lian:
Thanks for the issue report

Closing as upstream issue ref https://github.com/pydata/pandas-gbq/pull/143 

Please reopen if not upstream issue

PR #143, opened 2018-03-04
bump google auth version

maxim-lian:
Closes https://github.com/pydata/pandas-gbq/issues/73

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,8 +18,8 @@ def readme():
 
 INSTALL_REQUIRES = [
     'pandas',
-    'google-auth>=1.4.1',
-    'google-auth-oauthlib>=0.0.1',
+    'google-auth',
+    'google-auth-oauthlib',
     'google-cloud-bigquery>=0.29.0',
 ]
 
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ def readme():
 
 INSTALL_REQUIRES = [
     'pandas',
-    'google-auth>=1.0.0',
+    'google-auth>=1.0.2',
     'google-auth-oauthlib>=0.0.1',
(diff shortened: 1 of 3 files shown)


parthea:
Could we upgrade to the latest version of `google-auth` to take advantage of the latest fixes? The latest version 1.4.1 might also fix #136 . I see a note in the `google-auth` [release notes](https://github.com/GoogleCloudPlatform/google-auth-library-python/releases) that says `'default() now checks for the project ID environment var before warning about missing project ID.'`

maxim-lian:
Yes, to be clear, we want to make that a _min_ version?

maxim-lian:
@parthea gentle ping

setup.py
```diff
@@ -18,7 +18,7 @@ def readme():
 
 INSTALL_REQUIRES = [
     'pandas',
-    'google-auth>=1.0.0',
+    'google-auth>=1.4.1',
```
tswast: Hmm, I'm not so sure about bumping up a dependency version here just for upstream bug fixes. This is technically a breaking change and I know it would cause some hurt inside Google, which is ironically slow to make version upgrades of these packages.

We should keep our CI and this in sync, though. What was the reason for bumping the version to 1.0.2 in https://github.com/pydata/pandas-gbq/pull/139 ?

parthea: @tswast @maxim-lian
How do you feel about removing pinning altogether ?

tswast: I'd be more okay with that.

maxim-lian: OK - let's do that

I _think_ we can then refer those issues (#73 and #136) upstream - they're not pandas-gbq specific. Is that right?

maxim-lian:
Updated!

PR #147, opened 2018-03-10
TST: Fix Unit Tests

maxim-lian:
- Fixed most of these (+ parametrized them). Two remaining are xfailed. I'm not sure whether they should be passing?
- Removed a test testing old dependencies

### DIFF
--- a/pandas_gbq/tests/test_gbq.py
+++ b/pandas_gbq/tests/test_gbq.py
@@ -280,25 +280,7 @@ class TestGBQConnectorIntegrationWithLocalUserAccountAuth(object):
         assert isinstance(credentials, Credentials)
 
 
-@pytest.mark.skip('Currently fails, see '
-                  'https://github.com/pydata/pandas-gbq/pull/125')
 class TestGBQUnit(object):
-
-    def test_import_google_api_python_client(self):
-        if not _in_travis_environment():
-            pytest.skip("Skip if not in travis environment. Extra test to "
-                        "make sure pandas_gbq doesn't break when "
-                        "using google-api-python-client==1.2")
-
-        if compat.PY2:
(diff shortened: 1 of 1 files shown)


maxim-lian:
Gentle ping @tswast @parthea

pandas_gbq/tests/test_gbq.py
```diff
@@ -280,53 +280,31 @@ def test_get_user_account_credentials_returns_credentials(self):
...
-        assert not result
-
-    def test_should_return_bigquery_strings_as_python_strings(self):
-        result = gbq._parse_entry('STRING', 'STRING')
-        assert result == 'STRING'
+    @pytest.mark.parametrize(
+        ('input', 'type_', 'expected'), [
+            (1, 'INTEGER', int(1)),
+            (1, 'FLOAT', float(1)),
+            pytest.param('false', 'BOOLEAN', False, marks=pytest.mark.xfail),
```
tswast: Interesting that this was here before. AFAIK BigQuery returns booleans as proper booleans in the JSON returned. Probably a case of `pandas-gbq` just being extra conservative in what it expects?

pandas_gbq/tests/test_gbq.py
```diff
@@ -280,53 +280,31 @@ def test_get_user_account_credentials_returns_credentials(self):
...
+            pytest.param(
+                '0e9', 'TIMESTAMP',
```
tswast: Does this work if 0 is passed in as a float and not passed as a string?

maxim-lian:
Thanks for merging @tswast  - I was waiting to answer those questions - but you're right better to get it in and come back to them

tswast:
Yeah, I think having unit tests in `master` is more important than solving those right now.

PR #150, opened 2018-03-14
Load data with user defined schema

aktech:
The `to_gbp` function fails when there is a user-defined schema which is slightly different from the schema of DataFrame.

Traceback:

```
  File "<stdin>", line 4, in <module>
  File "pandas_gbq/gbq.py", line 979, in to_gbq
    schema=table_schema)
  File "pandas_gbq/gbq.py", line 569, in load_data
    self.process_http_error(ex)
  File "pandas_gbq/gbq.py", line 450, in process_http_error
pandas_gbq.gbq.GenericGBQException: Reason: 400 POST https://www.googleapis.com/upload/bigquery/v2/projects/aktech-labs/jobs?uploadType=resumable: Provided Schema does not match Table aktech-labs:pandas_test.gamma. Field A has changed type from FLOAT to STRING
```

The primary reason for this is that the `load_data` function ignores the `schema` argument.

Solution: The `schema` argument is now passed to the `load_chunks` function inside `load_data` function. A suitable integration test has been added.

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -562,7 +562,7 @@ class GbqConnector(object):
         try:
             for remaining_rows in _load.load_chunks(
                     self.client, dataframe, dataset_id, table_id,
-                    chunksize=chunksize):
+                    chunksize=chunksize, schema=schema):
                 self._print("\rLoad is {0}% Complete".format(
                     ((total_rows - remaining_rows) * 100) / total_rows))
         except self.http_error as ex:
--- a/pandas_gbq/tests/test_gbq.py
+++ b/pandas_gbq/tests/test_gbq.py
@@ -1412,6 +1412,23 @@ class TestToGBQIntegration(object):
         result = result_df['times'].sort_values()
         tm.assert_numpy_array_equal(expected.values, result.values)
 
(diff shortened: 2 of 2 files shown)


maxim-lian:
Superseeds https://github.com/pydata/pandas-gbq/pull/131

maxim-lian:
This looks great. Any feedback @tswast ? 

@aktech did you run the tests? Otherwise I will. (Travis integration tests don't work on PRs because of the auth keys)

@tswast more generally, do you think it's OK to merge these small PRs that look good, so the tests run on master, and then correct in the 5% of cases where it fails? Or we should run the tests on an 'auth-enabled' travis first?

tswast:
LGTM.

@maxim-lian So long as whomever merges keeps an eye on the master build and fixes / reverts based on the outcome, I'm okay merging without requiring an auth-enabled Travis.

aktech:
Thanks for the quick reply and merging.

> @aktech did you run the tests? Otherwise I will. (Travis integration tests don't work on PRs because of the auth keys)

Yes, I did ran the tests locally as well as on travis:
https://travis-ci.org/aktech/pandas-gbq/builds/353407949

PR #152, opened 2018-03-22
BUG: Adapt to breaking change in google-cloud-bigquery 0.32.0.dev1

tswast:
There was a breaking change in 0.32.0 which changed the way
configuration for the query job gets loaded. Also, it added the
'description' field to the schema resource, so this change updates the
schema comparison logic to account for that.

Note: I've created this PR based on the code currently in `master` in the [google-cloud-bigquery repo](https://github.com/GoogleCloudPlatform/google-cloud-python/tree/master/bigquery). The tests pass locally, but they won't pass anywhere else until google-cloud-bigquery 0.32.0 is released. I've added the **DO NOT MERGE** label to this PR for now.

Since pandas-gbq will be broken when google-cloud-bigquery does its next release, I think it would make sense to review this before than so it can be merged when everything is ready. Once it goes in, I can create a pandas-gbq release to minimize the breakage time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ install:
       conda install -q numpy pytz python-dateutil;
       PRE_WHEELS="https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com";
       pip install --pre --upgrade --timeout=60 -f $PRE_WHEELS pandas;
+      pip install -e 'git+https://github.com/GoogleCloudPlatform/google-cloud-python.git#egg=version_subpkg&subdirectory=api_core';
+      pip install -e 'git+https://github.com/GoogleCloudPlatform/google-cloud-python.git#egg=version_subpkg&subdirectory=core';
+      pip install -e 'git+https://github.com/GoogleCloudPlatform/google-cloud-python.git#egg=version_subpkg&subdirectory=bigquery';
     else
       conda install -q pandas=$PANDAS;
     fi
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -4,7 +4,6 @@ import os
 import time
 import warnings
 from datetime import datetime
-from distutils.version import StrictVersion
 from time import sleep
 
 import numpy as np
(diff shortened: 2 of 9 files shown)


pandas_gbq/gbq.py
```diff
@@ -627,17 +630,9 @@ def verify_schema(self, dataset_id, table_id, schema):
...
-        # Ignore mode when comparing schemas.
-        for field in fields_local:
-            if 'mode' in field:
-                del field['mode']
-        for field in fields_remote:
-            if 'mode' in field:
-                del field['mode']
+        fields_remote = self._clean_schema_fields(
+            self.schema(dataset_id, table_id))
+        fields_local = self._clean_schema_fields(schema['fields'])
```
maxim-lian: much nicer

maxim-lian:
I think we should either make this backwards compatible or check the version on import and raise. Otherwise people are going to get very obscure error messages if their imports aren't correct (which is very possible with pinned installs)

tswast:
Agreed. I had a hard time figuring out how to make this one backwards compatible since old versions would accept the config but then the API call to BigQuery fails later on.

I've updated the version number in `_check_google_client_version()`, so folks should get an understandable error message.

maxim-lian:
Though IIUC it's only backward incompatible for those passing in config dicts, so a very small share of users. Not worth a big lift, then

maxim-lian:
> I've updated the version number in _check_google_client_version(), so folks should get an understandable error message.

Ah great! (some crossed wires with my last msg)

pandas_gbq/gbq.py
```diff
@@ -463,28 +463,22 @@ def run_query(self, query, **kwargs):
...
-                raise ValueError("Only 'query' job type is supported")
+            job_config.update(config)
+
+            if 'query' in config and 'query' in config['query']:
+                if query is not None:
+                    raise ValueError("Query statement can't be specified "
+                                     "inside config while it is specified "
+                                     "as parameter")
+                query = config['query']['query']
+                del config['query']['query']
```
maxim-lian: no worth changing, but these two could be `query = config['query'].pop('query')`

maxim-lian:
We could update the final build to also use master from `google.cloud.bigquery` as well as pandas - would be generally helpful as well as for this

tswast:
@maxim-lian Good idea. I've made the "MASTER" build use the version of google-cloud-python hosted on GitHub. https://github.com/pydata/pandas-gbq/pull/152/commits/5f427a02c675c4a010761e306e2becbc7ae3385c

I'll keep an eye on the MASTER build on Travis to see if it works. https://travis-ci.org/tswast/pandas-gbq/builds/357586225

.travis.yml
```diff
@@ -28,6 +28,7 @@ install:
       conda install -q numpy pytz python-dateutil;
       PRE_WHEELS="https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com";
       pip install --pre --upgrade --timeout=60 -f $PRE_WHEELS pandas;
+      pip install -e 'git+https://github.com/GoogleCloudPlatform/google-cloud-python.git#egg=version_subpkg&subdirectory=bigquery';
```
maxim-lian: I've struggled with these from time to time. I _think_ the `egg=version_subpkg` is like `foo` and can be excluded. But if this works then great

tswast: Probably. I don't really want to mess with it, though. I got it from this [answer on installing with pip from a git subdirectory](https://stackoverflow.com/a/19516714/101923).

tswast:
Finally got a green build on MASTER. https://travis-ci.org/tswast/pandas-gbq/jobs/357606308 Got failures on the other builds as expected, since google-cloud-bigquery hasn't released 0.32.0 yet.

.travis.yml
maxim-lian: It works!

tswast:
@maxim-lian I figured out how to make it backwards compatible so that we don't have to force upgrade from 0.29.0 versions. Basically, I use the same version detection code we're already doing in the `_check_google_client_version()` function but in a new function to construct the query configuration.

Please take a look. With this new way of doing it, I think we can merge without waiting to sync with a release of `google-cloud-bigquery`.

tswast:
New Travis build: https://travis-ci.org/tswast/pandas-gbq/builds/361769406

maxim-lian:
Very nice! Good job on the backward-compat.

maxim-lian:
Re `_query.py` - I haven't seen other libraries that use private module names. It's fine and not worth changing, but generally libraries put the public API in `pandas_gbq.__init__`, and then there's no need to make the module names private.

tswast:
> generally libraries put the public API in pandas_gbq.__init__, and then there's no need to make the module names private.

Makes sense. Filed https://github.com/pydata/pandas-gbq/issues/154 to follow-up with some module renames.

Issue #52, opened 2017-06-07
TST: Add support for TOX


parthea:
It could be helpful to setup `tox` for this repo. Once it is setup simply running `tox` at the root of the repo will run tests in both versions, as well as flake8.

tswast:
I added this in #160. I added some light instructions for using it at https://github.com/pydata/pandas-gbq/blob/master/docs/source/contributing.rst#testing-on-multiple-python-versions

PR #160, opened 2018-04-06
TST: use nox for testing

tswast:
I couldn't actually reproduce the test failures introduced in https://github.com/pydata/pandas-gbq/pull/158 locally. My hope is that by using `nox` (like `tox`, but configurable via Python code) the more isolated environments will fix whatever error with mocks we're getting.

A step towards https://github.com/pydata/pandas-gbq/issues/52.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ dist
 **/wheelhouse/*
 # coverage
 .coverage
+.nox
 
 # OS generated files #
 ######################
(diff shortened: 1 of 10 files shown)


tswast:
@maxim-lian I think I finally fixed the busted tests. 😤 https://travis-ci.org/pydata/pandas-gbq/builds/363360912 Full integration tests still running at https://travis-ci.org/tswast/pandas-gbq/builds/363360902

I think the issue before was that there is some conflict between installing packages via pip and installing packages via conda. Somehow a bad version of the `mock` module ended up getting installed.

The new setup uses `nox` (similar to `tox`) for `pip`-based installs and keeps the previous logic for `conda`-based installs.

PR #163, opened 2018-04-10
CLN: rename internal modules. move tests.

### LINKED ISSUE #154: CLN: No need to use private module names.
Anything in `pandas_gbq.__init__` is considered public. Everything else should be considered private. No need to put underscore before the private module names.

See: https://github.com/pydata/pandas-gbq/pull/152#issuecomment-378379342

We could do a refactor PR to rename the existing modules and test modules.

tswast:
Per GH#154 we don't need to use underscore to show submodules are
private. To be 100% clear, I have added a disclaimer to the API
reference as well.

Also, I have split the system tests from the unit tests so that the
(fast) unit tests can be more easily run separately from the (slow)
system tests. I have followed the same directory structure as used in
the google-cloud-bigquery library.

Closes #154


maxim-lian:
This is great! Thanks a lot.

(and apologies for being offline for a few days - I owe you a response on the other PR)

tswast:
Tests running at https://travis-ci.org/tswast/pandas-gbq/builds/364840327

I need to investigate a local failure of
`tests/system.py::TestReadGBQIntegration::test_bad_project_id[local] FAILED` but curiously not with
```
tests/system.py::TestReadGBQIntegration::test_bad_project_id[service_path] PASSED                                                                     [ 38%]
tests/system.py::TestReadGBQIntegration::test_bad_project_id[service_creds] PASSED
```, but that feels like it might be more my config than an actual failure, since all the other local auth tests pass.

maxim-lian:
I see an issue with `UPDATING build/lib/pandas_gbq/_version.py`?

tswast:
Good catch. I'm seeing more references to `_version.py` than I first thought: `.gitattributes`, `MANIFEST.in`, `setup.cfg`, and even within `_version.py` itself. It looks like [Versioneer expects a `_version.py` file](https://github.com/warner/python-versioneer/blob/master/INSTALL.md) so I've moved that one back.

tswast:
It's working now: https://travis-ci.org/tswast/pandas-gbq/builds/364862859

Issue #162, opened 2018-04-07
Progress bar for to_gbq


cgarciae:
It would be awesome to get some feed back on the status of the upload if possible.

maxim-lian:
👍 

PRs welcome here

PR #166, opened 2018-04-19
Add progress for to_gbq function

aktech:
Not sure, if we are fine with using a library (`tqdm`) or we would like to create our own progress bar.

TODO:

- [ ] Run tests locally


aktech:
Thanks for the quick review @maxim-lian I will make the changes soon.

requirements.txt
```diff
@@ -2,3 +2,4 @@ pandas
 google-auth
 google-auth-oauthlib
 google-cloud-bigquery
+tqdm
```
tswast: requirements.txt is only used by developers / CI testing. You need to add this to `setup.py` for it to actually get listed as a dependency.

As @maxim-lian says, we could make this an optional dependency. If you go that route, use the ["extras" section of setup.py](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) ([example](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/1040e8b00eb63fe1dda38094cb4fe9c6efe55f04/bigquery/setup.py#L88).

pandas_gbq/gbq.py
```diff
@@ -551,16 +551,20 @@ def run_query(self, query, **kwargs):
...
         try:
-            for remaining_rows in load.load_chunks(
-                    self.client, dataframe, dataset_id, table_id,
-                    chunksize=chunksize, schema=schema):
+            chunks = load.load_chunks(self.client, dataframe, dataset_id,
+                                      table_id, chunksize=chunksize,
+                                      schema=schema)
+            if progress_bar:
+                from tqdm import tqdm
+                chunks = tqdm(chunks)
```
tswast: In many cases there will be just one chunk, even for quite large dataframes. To be a useful progress indicator, this change should manually poll to see if the load job is complete as is done for query jobs:

https://github.com/pydata/pandas-gbq/blob/8f49ec3f134cc0a85f541116c42647c515fdc7e6/pandas_gbq/gbq.py#L504-L522

maxim-lian:
I just tried tqdm locally and it's v good - much nicer than the current logging. I vote that we at least enable-by-default-if-installed.

aktech:
> I just tried tqdm locally and it's v good - much nicer than the current logging. I vote that we at least enable-by-default-if-installed.

Indeed.

pandas_gbq/gbq.py
aktech: Well, tqdm can to anything in this case. It's more useful for more number of chunks. The percentage done is based on the proportion of of chunks completed wrt the total number of chunks.

pandas_gbq/gbq.py
```diff
@@ -551,23 +551,34 @@ def run_query(self, query, **kwargs):
...
                 logger.info("\rLoad is {0}% Complete".format(
                     ((total_rows - remaining_rows) * 100) / total_rows))
         except self.http_error as ex:
             self.process_http_error(ex)
 
         logger.info("\n")
 
+    def _check_if_tqdm_exists(self):
+        try:
+            import tqdm  # noqa
```
tswast: Since this dependency is optional, there's no need to put it in a function. The only reason the `google-cloud-bigquery` imports are in a function is that they are required and it helps to make a better error message. Ideally there should be no errors if tqdm is not installed.

Example of optional import:

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/ecb501c1d8b099af1d804ad1cfd2cb96575faf3e/bigquery/google/cloud/bigquery/table.py#L24-L27

pandas_gbq/gbq.py
tswast: I'd caution against this change in that case. Load jobs are a limited resource (1,000 per day), so I'd rather encourage people to use fewer load jobs than more. Chunks should only be used when there is extreme memory pressure and a dataframe does not fit into memory if we try to serialize to CSV before upload.

pandas_gbq/gbq.py
```diff
@@ -551,23 +551,34 @@ def run_query(self, query, **kwargs):
...
+                chunks = tqdm(chunks)
+            for remaining_rows in chunks:
                 logger.info("\rLoad is {0}% Complete".format(
                     ((total_rows - remaining_rows) * 100) / total_rows))
         except self.http_error as ex:
             self.process_http_error(ex)
 
         logger.info("\n")
 
+    def _check_if_tqdm_exists(self):
```
maxim-lian: I don't mean to slow you down with tweaks. That said, there's no need for this to be a method - you can have this inline like https://github.com/pandas-dev/pandas/blob/73085773998e12e85f1044771069f63f4e8d65ad/pandas/tests/generic/test_frame.py#L23

maxim-lian:
@tswast do you know whether these are failures from master rather than @aktech 's work? https://travis-ci.org/pydata/pandas-gbq/jobs/370512748

tswast:
@maxim-lian Those failures look like a bad version of `setuptools` is failing to install the `google-api-core` namespaced package. Probably unrelated to this change.

tswast:
I actually don't think this change addresses #162 for the reasons I state in  https://github.com/pydata/pandas-gbq/pull/166#discussion_r182808563

Since there is already a logging message for percentage based on chunks, I believe #162 is for the default (and encouraged) case when there is just one chunk. To show progress in the case of a single chunk, one must manually poll to see if the load job is complete as is done for query jobs:

https://github.com/pydata/pandas-gbq/blob/8f49ec3f134cc0a85f541116c42647c515fdc7e6/pandas_gbq/gbq.py#L504-L522

Why do I say a single chunk should be encouraged? Load jobs are a limited resource (1,000 per day), so I'd rather encourage people to use fewer load jobs than more. Chunks should only be used when there is extreme memory pressure and a dataframe does not fit into memory if we try to serialize to CSV before upload.

maxim-lian:
Right, I think you could still have use this for one chunk - though it would be a counter rather than a progress bar (unless anyone can find a way of getting actual progress?)

Something like:

```python

def wait_for_job(job, timeout_in_seconds=600):
    # https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/cloud-client/snippets.py
    start = datetime.datetime.now()
    timeout = start + datetime.timedelta(0, timeout_in_seconds)
    with tqdm(
            bar_format='Waiting for {desc} Elapsed: {elapsed}',
            total=10000) as progress:
        while True:
            progress.set_description(str(job))
            if job.state == 'DONE':
                if job.error_result:
                progress.bar_format = 'Completed {decs}. Elapsed: {elapsed}'
                return
            if datetime.datetime.now() > timeout:
                raise IOError
            time.sleep(1)
```

pandas_gbq/gbq.py
```diff
@@ -551,23 +551,34 @@ def run_query(self, query, **kwargs):
...
-                    self.client, dataframe, dataset_id, table_id,
-                    chunksize=chunksize, schema=schema):
+            chunks = load.load_chunks(self.client, dataframe, dataset_id,
+                                      table_id, chunksize=chunksize,
+                                      schema=schema)
+            if progress_bar and self._check_if_tqdm_exists():
+                from tqdm import tqdm
+                chunks = tqdm(chunks)
+            for remaining_rows in chunks:
                 logger.info("\rLoad is {0}% Complete".format(
```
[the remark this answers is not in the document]
aktech: Yeah, indeed. I forgot to remove that.

maxim-lian:
Looks like GH thought I was still requesting changes. I just clicked the button. 

@aktech we'll merge now and then can add any improvements in the future?

aktech:
@maxim-lian @tswast sure.

pandas_gbq/gbq.py
```diff
@@ -563,16 +568,19 @@ def run_query(self, query, **kwargs):
...
-            for remaining_rows in load.load_chunks(
-                    self.client, dataframe, dataset_id, table_id,
-                    chunksize=chunksize, schema=schema):
+            chunks = load.load_chunks(self.client, dataframe, dataset_id,
+                                      table_id, chunksize=chunksize,
+                                      schema=schema)
+            if progress_bar and tqdm:
+                chunks = tqdm.tqdm(chunks)
+            for remaining_rows in chunks:
                 logger.info("\rLoad is {0}% Complete".format(
```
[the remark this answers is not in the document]
aktech: Yes, we can do it in another PR.

PR #170, opened 2018-04-29
Explicit auth tests

maxim-lian:
After I neglected https://github.com/pydata/pandas-gbq/pull/127 and @tswast valiantly completed the bulk, there were a couple of useful unrelated pieces that are also worth adding

Here, tests are explicitly marked with auth type so, for example, if `pytest -m 'not s_path_auth'` is run, it won't run any tests with service path auth
Travis is then parameterized by the command it's passed, rather than the tests themselves checking whether they're running in travis


nox.py
```diff
@@ -21,6 +20,12 @@ def default(session):
         'pytest',
         os.path.join('.', 'tests', 'unit'),
         os.path.join('.', 'tests', 'system.py'),
+        os.path.join('.', 'tests', 'unit'),
```
tswast: Looks like unit tests got included twice.

tests/system.py
```diff
@@ -205,6 +220,7 @@ def test_should_be_able_to_get_results_from_query(self, gbq_connector):
         assert pages is not None
 
 
+@pytest.mark.local_auth
```
tswast: It seems like `test_get_application_default_credentials_does_not_throw_error` and `test_get_application_default_credentials_returns_credentials` are working just fine on Travis if you want to move this to the `test_get_user_account_credentials_XXX` methods.

tswast:
This doesn't seem to be skipping local auth tests no matter where I put the mark.  :-(

```
$ nox -s default -r | grep TestAuth
nox > Running session default
nox > Re-using existing virtualenv at /Users/swast/src/pandas/pandas-gbq/.nox/default.
nox > chdir /Users/swast/src/pandas/pandas-gbq
nox > Skipping installation of "mock pytest pytest-cov".
nox > Skipping installation of "-e .".
nox > pytest ./tests/unit ./tests/system.py -m 'not local_auth and not s_path_auth' -v --cov=pandas_gbq --cov=tests.unit --cov-report xml:/tmp/pytest-cov.xml
tests/system.py::TestAuth::test_get_application_default_credentials_does_not_throw_error[env] PASSED [ 33%]
tests/system.py::TestAuth::test_get_application_default_credentials_returns_credentials[env] PASSED [ 33%]
tests/system.py::TestAuth::test_get_user_account_credentials_bad_file_returns_credentials[env] PASSED [ 34%]
tests/system.py::TestAuth::test_get_user_account_credentials_returns_credentials[env] PASSED [ 35%]
^Cnox > Interrupted...
nox > Session default interrupted.
```

maxim-lian:
Discovery, though not solution - if there's a legacy-style `autouse=True` fixture, even the explicit marking (like you added) doesn't work.

We could try and resolve, or move to new-style tests (which would be preferable, but non-zero work)

tswast:
Ugh. Okay, well I've wanted to move to new-style tests for a while. (Already did it with the unit tests.) I'll see if I can get a PR to refactor the system tests out today.

tswast:
Well, I only got to the auth tests, but it's a start. https://github.com/pydata/pandas-gbq/pull/176

tswast:
Closing in favor of #176 changes.

Issue #107, opened 2018-01-10
Request: add schemaUpdateOptions to to_gbq()


yiga2:
Useful if table already exists but `df` has more columns e.g. appending to historical data with 'old' schema.

Note that currently the (Python) Client library does not support that property - as per 0.29.

Workaround is `LoadJobConfig._properties.__setitem__('schemaUpdateOptions', ['ALLOW_FIELD_ADDITION'])`.

tswast:
Now that `to_gbq` is creating load jobs, it would be good to allow passing in a whole job config dictionary in the way that `read_gbq` does with it’s `configuration` argument.

tswast:
Closing in favor of https://github.com/googleapis/python-bigquery-pandas/issues/425

PR #173, opened 2018-05-04
CLN: remove workaround for streaming delays.

tswast:
Now that load jobs are used, no sleep is required when replacing a table.

Tests with credentials running at https://travis-ci.org/tswast/pandas-gbq/builds/375003060

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -4,7 +4,6 @@ import os
 import time
 import warnings
 from datetime import datetime
-from time import sleep
 
 import numpy as np
 from pandas import DataFrame, compat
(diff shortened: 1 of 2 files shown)


maxim-lian:
Nice.

Slight tangent, but we could think about only doing one chunk, which would mean we could fall back to the replace / amend / overwrite native implementations rather than writing our own

tswast:
Yeah, someone reported that at https://github.com/pydata/pandas-gbq/issues/107 I agree we should do it. We could even do it if we keep chunking by having subsequent load jobs be "append" mode, but it could also be a good opportunity to deprecate chunks since it's actually not all that helpful when using load jobs (which are a limited resource of 1,000 / day).

PR #176, opened 2018-05-05
CLN: refactor auth logic to its own module.

tswast:
Uses new-style tests for auth-related tests. Also adds a couple additional auth-related tests for better coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ script:
   - if [[ $PYTHON == '3.6' ]] && [[ "$PANDAS" == "MASTER" ]]; then nox -s test36master ; fi
   - REQ="ci/requirements-${PYTHON}-${PANDAS}" ;
     if [ -f "$REQ.conda" ]; then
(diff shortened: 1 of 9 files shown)


tswast:
Travis build with credentials at https://travis-ci.org/tswast/pandas-gbq/builds/375140209

tswast:
Oops. Looks like I need to update some of the monkey patches we used to skip auth in the unit tests and move over a few more tests from system/test_gbq.py to system/test_auth.py

tswast:
Note: I borrowed the test marking idea for `local_auth` from #170. For the other kinds of credentials, I automatically skip in the fixture when the credentials can't be fetched.

Since there is more coverage for the other kinds of auth now, I think the rest of #170 might be unnecessary?

maxim-lian:
Great!! Thanks so much for pushing all the way through on these. Both the main code and the tests are better.

Yes, I think we can close #170 now.

PR #184, opened 2018-05-25
TST: Add unit tests for pandas_gbq.auth.get_credentials().

### LINKED ISSUE #49: TST: Improve unit tests for GoogleCredentials.get_application_default()
The test `test_get_application_default_credentials_returns_credentials()` : https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/tests/test_gbq.py#L297 is skipped if `_check_if_can_get_correct_default_credentials()` returns `false`. This happens when default credentials are not available. In order to improve test coverage, specifically on Travis, we could (temporarily) populate the default credentials during testing in order to exercise the code that uses default credentials. We could unset the default credentials after testing.

tswast:
Tests (actual auth mocked out):

* Using private key with contents.
* Using private key with path.
* Using default credentials.
* Using cached user credentials.

Also, since I found it a little hard to debug the private key tests, update the error message with a little more information about what kind of credentials it thinks `private_key` is.

Closes #49

### DIFF
--- a/pandas_gbq/auth.py
+++ b/pandas_gbq/auth.py
@@ -39,8 +39,10 @@ def get_service_account_credentials(private_key):
     import google.auth.transport.requests
     from google.oauth2.service_account import Credentials
 
+    is_path = os.path.isfile(private_key)
+
     try:
-        if os.path.isfile(private_key):
+        if is_path:
             with open(private_key) as f:
                 json_key = json.loads(f.read())
         else:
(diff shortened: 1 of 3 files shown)


tswast:
Failing on Python 2.7. I'm investigating.

tswast:
It's a mock issue. The Python 2.7 autospec seems to be getting confused by the use of a classmethod.

PR #187, opened 2018-06-11
TST: Lock miniconda version

tswast:
The latest version of miniconda seems to be having a hard time with the `google.cloud` and `google.api_core` namespaced packages again. Hopefully reverting back to an old version of miniconda fixes it.


tswast:
Still fails to load the `google.api_core` package. https://travis-ci.org/pydata/pandas-gbq/jobs/390816271

Issue #161, opened 2018-04-07
pandas-gbq auth proposal


tswast:
## Overview

The current auth flows for pandas-gbq are a bit confusing and hard to customize. 

**Final desired state**. The `pandas_gbq` module should have the following (changes in **bold**):

*   read_gbq(query, project_id [**optional**], index_col=None, col_order=None, reauth, verbose [deprecated], private_key [**deprecated**], auth_local_webserver, dialect='legacy', configuration [optional], credentials [**new param, optional**])
*   to_gbq(dataframe, destination_table, project_id [**optional**], chunksize=None, verbose [deprecated], reauth, if_exists='fail', private_key [**deprecated**], auth_local_webserver, table_schema=None, credentials [**new param, optional**])
*   **CredentialsCache** (and **WriteOnlyCredentialsCache**, **NoopCredentialsCache**) - new class (and subclasses) for configuring user credentials caching behavior
*   **context** - global singleton with "client" property for caching default client in-memory.
*   **get_user_credentials**(**scopes**=None, **credentials_cache**=None, **client_secrets**=None, **use_localhost_webserver**=False) - Helper function to get user authentication credentials.

**Tasks**:

*   [x] Add authentication documentation with examples.
*   [x] Add optional credentials parameter to `read_gbq`, taking a google.cloud.bigquery.Client object.
*   [x] Add optional credentials parameter to `to_gbq`, taking a google.cloud.bigquery.Client object.
*   [x] Add pandas_gbq.get_user_credentials() helper for fetching user credentials with installed-app OAuth2 flow.
*   [x] Add pandas_gbq.CredentialsCache and related subclasses for managing user credentials cache.
*   [x] Add pandas_gbq.context global for caching a default Client in-memory. Add examples for manually setting pandas_gbq.context.client (so that default project and other values like location can be set).
*   [x] Update minimum google-cloud-bigquery version to 0.32.0 so that the project ID in the client can be overridden when creating query & load jobs. (Done in https://github.com/pydata/pandas-gbq/pull/185)
*   [x] Deprecate `private_key` argument. Show examples of how to do the same thing by passing  [Credentials](https://google-auth.readthedocs.io/en/latest/reference/google.auth.credentials.html#google.auth.credentials.Credentials)  to the Client constructor.
*   [x] Deprecate PANDAS_GBQ_CREDENTIALS_FILE environment variable. Show example using `pandas_gbq.get_user_credentials` with `credentials_cache` argument.
[...]

Issue #149, opened 2018-03-13
document pandas-gbq vision and roadmap


max-sixty:
Both pandas-gbq and [google-cloud-bigquery](https://github.com/GoogleCloudPlatform/google-cloud-python/tree/master/bigquery) are doing many of the same things, and increasingly so (e.g. `.to_dataframe()` in google-cloud-bigquery)

- Are there different use cases? Can we define those?
- Should we focus development on one and wrap the other? Even if not wholly, for a subset of functionality?
- Is there some direction from Google? @tswast spends a lot of time on both libraries so he is probably best placed to offer guidance

tswast:
Yeah, there's overlap. There are some things that `pandas-gbq` handles well that `google-cloud-bigquery` doesn't (or at all).

Not supported at all in `google-cloud-bigquery`:
- Uploading a dataframe to a BigQuery table. I would consider this part of the feature request at https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4553
- Built-in user-based authentication (3-legged OAuth, 3LO). Based on conversations with @jonparrott, I think it's probably never that `google-cloud-bigquery` would support this. It would be inconsistent with other products, even though I think it would make sense for BigQuery.

Not supported well in `google-cloud-bigquery`:
- Dataframes with multiple indexes.
- Probably missing some of the bug fixes that have gone into `pandas-gbq`.

In `google-cloud-bigquery` but not in `pandas-gbq`:
- IPython magics (just added in a recent PR, but not released).

**The future?** My thought is that more and more `pandas-gbq` becomes a thin wrapper over `google-cloud-python`. For example,
- Make `pandas-gbq` use `to_dataframe()`. (I tried this but got many failing system tests due to problems with multiple index dataframes.)
- Make `pandas-gbq` automatically register the IPython magic if available.

My thought is that we'd always have `pandas-gbq` specifically for the user authentication piece. If I can come up with a user auth proposal that would be acceptable for `google-cloud-bigquery`, my opinion would change, but right now that seems unlikely.

tswast:
Some updates:

I filed https://github.com/pydata/pandas-gbq/issues/175 and https://github.com/pydata/pandas-gbq/issues/174 while investigating the `to_dataframe()` function. Both of these issues would be solved by moving to `google-cloud-bigquery`'s `to_dataframe()` method.

[...]

PR #191, opened 2018-06-26
Add anchor links to versions in the changelog

tswast:
Uses [Sphinx syntax for cross-reference labels](https://stackoverflow.com/a/19543591/101923).

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,8 @@
 Changelog
 =========
 
+.. _changelog-0.5.0:
+
 0.5.0 / 2018-06-15
 ------------------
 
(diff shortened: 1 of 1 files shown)


maxim-lian:
Thanks @tswast

tswast:
Of course!

For posterity, I added these anchors so that I could link to them from the Pandas docs at https://github.com/pandas-dev/pandas/pull/21628

maxim-lian:
Super!

I've been out of action on OSS recently given work, and heading on vacation for a couple of weeks. Hopefully I'll have more time in the Summer. Let me know if you see pandas-gbq issues that you think would be worth focusing on. Would be good for me to get something in re bulk download.

tswast:
No worries. These are the top priorities on my mind for pandas-gbq:

1. Auth. https://github.com/pydata/pandas-gbq/issues/161 Basically, add a `client` parameter plus a couple public helper methods for auth and deprecate all the other auth-related parameters.
2. `to_gbq` configuration argument. Take in a job configuration dictionary, just like `read_gbq` does. It might make sense to officially deprecate the chunksize argument at that time, too. https://github.com/pydata/pandas-gbq/pull/124#discussion_r169198218
3. Delegate more to `google-cloud-bigquery` https://github.com/pydata/pandas-gbq/issues/149#issuecomment-400459052

Bulk download will be an interesting one to investigate. It does depend on GCS, so it'll likely have to be a separate `read_gbq_bulk` method, since we need a scratch bucket to work with. (Plus we'll have to update the scopes we use for authentication.)

Issue #192, opened 2018-07-22
Writing to CSV buffer with default float format causes numerical overflow in BigQuery


anthonydelage:
I'm using `to_gbq()` to load a local DataFrame into BigQuery. I'm running into an issue where floating point numbers are gaining significant figures and therefore causing numerical overflow errors when loaded to BigQuery.

The `load.py` module's `encode_chunk()` function writes to a local CSV buffer using Pandas' `to_csv()` function, which has a known issue regarding added significant figures on some operating systems (read more [here](https://stackoverflow.com/questions/12877189/float64-with-pandas-to-csv)).

In my case, 0.208 was transformed to 0.20800000000000002.

I've been able to solve the issue locally by changing the `float_format` parameter to `'%g'` in the `encode_chunk()` function's `pd.to_csv()` [call](https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/load.py#L16):
```python
dataframe.to_csv(
    csv_buffer, index=False, header=False, encoding='utf-8',
    float_format='%g', date_format='%Y-%m-%d %H:%M:%S.%f')
```

Can this be safely applied as a default?

Versions:
```
pandas==0.22.0
pandas-gbq==0.5.0
```

OS details:
```
MacOS 10.13.4
```

max-imlian:
Thanks for the report @anthonydelage 

Without being an expert in floats, that looks very reasonable. How would you feel about doing a pull-request?

anthonydelage:
@max-imlian yes, it's in #193.

In it, I made one change to the proposal above: increasing the number of significant figures to 15 (decimal), the max safely allowed under IEEE-754 double formatting.

parthea:
Fixed in https://github.com/pydata/pandas-gbq/commit/993fe55f7e1f44a15b282b415284774310fc530c

PR #193, opened 2018-07-23
Use general float format when writing to CSV buffer to prevent numerical overload

anthonydelage:
## Summary

Based on issue #192.

The `load.py` module's `encode_chunk()` function writes to a local CSV buffer using Pandas' `to_csv()` function, which has a [known issue](https://stackoverflow.com/questions/12877189/float64-with-pandas-to-csv) regarding added significant figures on some operating systems. This can cause numerical overflow when loading the data into BigQuery.

This PR adds the `float_format='%.15g'` parameter to the `to_csv()` call.

## Details

This section provides a bit more detail behind the `%.15g` float format.

As per the Python 3 string format [documentation](https://docs.python.org/3.6/library/string.html#formatspec), the `g` (general) format acts as such:
> For a given precision p >= 1, this rounds the number to p significant digits and then formats the result in either fixed-point format or in scientific notation, depending on its magnitude.

BigQuery stores floats as [IEEE-754 doubles](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#floating-point-type). 15 is the maximum number of significant digits that can be used for a floating point number to guarantee that it does not overflow (according to [Wikipedia](https://en.wikipedia.org/wiki/Double-precision_floating-point_format)):
> If a decimal string with at most 15 significant digits is converted to IEEE 754 double-precision representation, and then converted back to a decimal string with the same number of digits, the final result should match the original string.

Hence, we choose the general format with at most 15 decimal digits of precision. It reduces rounding loss from pandas to BigQuery to a minimum and does not increase storage requirements since floats are being stored as IEEE-754 doubles, anyway.

### DIFF
--- a/tests/unit/test_load.py
+++ b/tests/unit/test_load.py
@@ -4,6 +4,7 @@ import numpy
 import pandas
 
 from pandas_gbq import load
+from io import StringIO
 
 
 def test_encode_chunk_with_unicode():
@@ -18,6 +19,20 @@ def test_encode_chunk_with_unicode():
     csv_bytes = csv_buffer.read()
     csv_string = csv_bytes.decode('utf-8')
     assert u'信用卡' in csv_string
+
+
+def test_encode_chunk_with_floats():
+    """Test that floats in a dataframe are encoded with at most 15 significant
+        figures.
+
+    See: https://github.com/pydata/pandas-gbq/issues/192
+    """
+    input_csv = StringIO('01/01/17 23:00,1.05148,1.05153,1.05148,1.05153,4')
+    df = pandas.read_csv(input_csv, header=None)
+    csv_buffer = load.encode_chunk(df)
+    csv_bytes = csv_buffer.read()
+    csv_string = csv_bytes.decode('utf-8')
+    assert '1.05153' in csv_string
 
 
 def test_encode_chunks_splits_dataframe():
@@ -36,3 +51,5 @@ def test_encode_chunks_with_chunksize_none():
     remaining, buffer = chunks[0]
     assert remaining == 0
     assert len(buffer.readlines()) == 6
+
+
--- a/tests/unit/test_load.py
+++ b/tests/unit/test_load.py
@@ -27,7 +27,7 @@ def test_encode_chunk_with_floats():
 
     See: https://github.com/pydata/pandas-gbq/issues/192
     """
-    input_csv = StringIO('01/01/17 23:00,1.05148,1.05153,1.05148,1.05153,4')
+    input_csv = StringIO(u'01/01/17 23:00,1.05148,1.05153,1.05148,1.05153,4')
     df = pandas.read_csv(input_csv, header=None)
     csv_buffer = load.encode_chunk(df)
     csv_bytes = csv_buffer.read()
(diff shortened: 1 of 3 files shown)


max-imlian:
Thanks a lot @anthonydelage 

Could you add a test? You could use the case you put in the issue

The error the the current test is unrelated

anthonydelage:
@max-imlian I added a test based on an example provided [here](https://github.com/pandas-dev/pandas/issues/16452).

One issue that I did notice while writing the test, and that I'm not sure how to deal with, is dealing with numbers that will overflow in BigQuery, e.g. `1e20` or `1e-20`. With or without the proposed string format update, overflowing values are written to the CSV buffer using exponential notation, which fails upon load to BigQuery because it's interpreted as a string. It's the kind of value that should either fail or be transformed to `+Inf` in BigQuery anyway, but that users won't be able to easily debug when they look at their load logs because it'll appear as a data type mismatch instead of a numeric overflow.

tests/unit/test_load.py
```diff
@@ -20,6 +21,20 @@ def test_encode_chunk_with_unicode():
     assert u'信用卡' in csv_string
 
 
+def test_encode_chunk_with_floats():
+    """Test that floats in a dataframe are encoded with at most 15 significant
+        figures.
+
+    See: https://github.com/pydata/pandas-gbq/issues/192
+    """
+    input_csv = StringIO('01/01/17 23:00,1.05148,1.05153,1.05148,1.05153,4')
```
max-imlian: For Py2 to pass, you need to put `StringIO(u'...')` to pass unicode into `StringIO` (you can see the test failure in Travis)

max-imlian:
Yes, good point. Do we get an overflow in BQ but not pandas because pandas can handle larger numbers?

(though we can deal with that separately, no need to delay this PR I think)

anthonydelage:
I think both systems theoretically use IEEE-754 (pandas uses numpy's [float64](https://docs.scipy.org/doc/numpy-1.13.0/user/basics.types.html
) under the hood), but somehow the binary -> decimal -> binary conversion flow is causing issues.

Ideally, we'd have one of the following:
- Pandas send the data to BigQuery without the CSV intermediary, maintaining its binary format
- BigQuery learns how to interpret numbers in exponential notation as floats

tswast:
> Pandas send the data to BigQuery without the CSV intermediary, maintaining its binary format

In `google-cloud-bigquery` we chose to serialize to Parquet. Eventually I'd like to move this library over to using [load_table_from_dataframe()](https://googlecloudplatform.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.client.Client.load_table_from_dataframe.html#google.cloud.bigquery.client.Client.load_table_from_dataframe) but I've hesitated because it means (a) an additional dependency on pyarrow and (b) there might be some subtle behavior differences (I know there are for `read_gbq()`, but haven't tested with `to_gbq()`).

max-imlian:
This looks good! 

@anthonydelage do you want to add a note & your name to the whatsnew?

anthonydelage:
@max-imlian are you referring to the [changelog.rst](https://github.com/pydata/pandas-gbq/blob/master/docs/source/changelog.rst) file?

max-imlian:
Yes! 

And you can add "by @anthonydelage " if you like. You can start a 0.5.1 (Unreleased) section.

anthonydelage:
Done!

max-imlian:
Thanks @anthonydelage !

PR #204, opened 2018-08-29
Formatting with black

max-sixty:
As an example of how the repo would look using https://github.com/ambv/black

I find this great for repos I'm starting. I'm not sure it's worth the tradeoff of merge conflicts for existing projects, though.

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,10 +1,4 @@
 linters:
-  flake8:
-    max-line-length: 79
+  black:
+    config: ./pyproject.toml
     fixer: true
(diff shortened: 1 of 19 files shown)


max-sixty:
OK great, I can add that to CI.

Any thoughts @parthea @jreback, given this is a v broad change?

parthea:
Sounds good to me. Thanks for adding this, @max-sixty !

max-sixty:
@tswast please could you take a final look before I merge? I changed Stickler to use black rather than running in Travis on nox

Issue #189, opened 2018-06-15
Travis build fails for Conda package


tswast:
See: https://travis-ci.org/pydata/pandas-gbq/jobs/392794382

Per https://github.com/conda-forge/staged-recipes/wiki/Namespace-packages, I think we'll probably need to create some empty `google-namespace` and `google-cloud-namespace` Conda packages.

Applying my previous fix of locking the miniconda version didn't work this time. (https://github.com/pydata/pandas-gbq/pull/187)

max-sixty:
Do other libraries face this? 

I've been off the `pandas-gbq` contribs for a while, but would be up for having a passing build - not a great encouragement for users & contributors I think. Happy to dive in

tswast:
I think the root cause is that the `google-cloud-bigquery` package lives in the `google.cloud` namespace.

I think the solution is to create a `google-namespace` and a `google-cloud-namespace` package https://github.com/conda-forge/staged-recipes/wiki/Namespace-packages but I'm somewhat frustrated that some Conda versions seem to work fine.

max-sixty:
I saw you did some work on airflow - do you know how they handle it? 

TBH this is the most unfamiliar area for me (I don't use conda). But happy to jump in if you're taking a pause from your recent heavy contribs!

tswast:
As I understand it, Airflow doesn't use Conda on CI. (I also doesn't test the GCP operators on CI 😭.)

tswast:
It'll probably be next week Friday before I get a good chunk of time to do pandas-gbq stuff if you wanted to take a stab at it.

max-sixty:
>  doesn't test the GCP operators on CI

😱 

Looks like some people are working on it here: https://github.com/conda-forge/airflow-feedstock/issues/15

tswast:
With CircleCI and the miniconda3 container, the packages are installing okay (that was the problem on Travis). Strangely, the pandas DataFrame is getting `None`s instead of `float.nan`s for NULL floating point values.

```
/opt/conda/envs/test-environment/lib/python3.6/site-packages/pandas/util/testing.py:1010: in assert_attr_equal
    left_attr, right_attr)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = 'Attributes', message = 'Attribute "dtype" are different'
[...]

PR #206, opened 2018-08-30
Test on Python37

max-sixty:
What do you think about skipping conda tests for the moment @tswast ? 

The test on 3.7 seems less controversial (probably should split these up, but both small PRs)


.travis.yml
```diff
@@ -8,12 +8,13 @@ matrix:
   - os: linux
     python: 3.5
     env: PYTHON=3.5 PANDAS=0.18.1 COVERAGE='true' LINT='false'
+  # TODO: fix conda issues around namespace packages
```
tswast: Please link to https://github.com/pydata/pandas-gbq/issues/189 in the comment.

max-sixty:
@tswast let me know if you recognize these ADC errors in the 3.7 build: https://travis-ci.org/max-sixty/pandas-gbq/jobs/422752133

Could they be related to us using sudo (which is required for 3.7 it seems)?

tswast:
> http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/?recursive=true from the Google Compute Enginemetadata service. Status: 404 Response

I think this is related to the fact that Travis CI runs on GCE but it locks down access to APIs. Maybe `google-auth` doesn't account for this case?

tswast:
Python 3.7 builds added in https://github.com/pydata/pandas-gbq/pull/232

PR #212, opened 2018-09-08
[BUG] Don't load credentials from disk if reauth is True

tswast:
If the credentials on disk are invalid, then `get_user_credentials()` may fail
before it can fetch fresh credentials.

Noticed first and fixed in pydata-google-auth package in
https://github.com/pydata/pydata-google-auth/commit/7ce7f3f822fbaea7cc73878f52c61f8b38109f36

Note about pydata-google-auth: eventually I'd like to move pandas-gbq
to use that package for auth, but since I'm not sure when I'll have that
package ready, I'm patching pandas-gbq, too.


max-sixty:
> eventually I'd like to move pandas-gbq to use that package for auth

👍

Issue #211, opened 2018-09-05
TST: update to latest version of Nox


tswast:
https://github.com/theacodes/nox/releases/tag/2018.8.23

The biggest change relevant to us is that nox changed how to specify Python versions for sessions.

> Please consult the documentation on how to use @nox.session(python=[...]) to configure virtualenvs for sessions.

Relevant docs: https://nox.readthedocs.io/en/stable/config.html#virtualenv-config

PR #222, opened 2018-09-21
Use latest version of Nox for tests.

tswast:
Also, update lint session to use black to check.

Closes https://github.com/pydata/pandas-gbq/issues/211


max-sixty:
Thanks!

Issue #174, opened 2018-05-04
pandas-gbq handles nulls in numeric columns differently from pandas


tswast:
[Pandas encodes missing / null data as NaN in numeric columns](https://pandas.pydata.org/pandas-docs/stable/missing_data.html).

pandas-gbq expects the type of a column containing Nulls to be object.
https://github.com/pydata/pandas-gbq/blob/f301442082bab62c793b6a80cf00c03f97938609/tests/system.py#L295-L302

Shouldn't pandas-gbq align with the choice of pandas in this case?

tswast:
Note that for datetime/timestamp columns the pandas-gbq behavior is a aligned with [pandas by using NaT for missing dates](https://pandas.pydata.org/pandas-docs/stable/missing_data.html#datetimes).

https://github.com/pydata/pandas-gbq/blob/f301442082bab62c793b6a80cf00c03f97938609/tests/system.py#L383-L387

maxim-lian:
> Shouldn't pandas-gbq align with the choice of pandas in this case?

You do then have all ints as floats. If you want floats with `NaN`s, you can cast to float in the query. And with the change there's no such escape hatch.

But I'm +0.2 to make the change - it's far more likely that someone would want a float column than an object column.

tswast:
> You do then have all ints as floats.

Yeah, I find that a bit odd, but it's what the DataFrame constructor does in the case of missing values.

```
In [1]: import pandas as pd

In [2]: df_int = pd.DataFrame([[1], [2], [3]])

In [3]: df_int.dtypes
Out[3]:
0    int64
dtype: object

In [4]: df_int_null = pd.DataFrame([[1], [None], [3]])

In [5]: df_int_null.dtypes
Out[5]:
0    float64
dtype: object
```

max-sixty:
Closed by #224

msegado:
Quick FYI, it looks like the docs still describe the old behavior:
https://pandas-gbq.readthedocs.io/en/stable/intro.html#integer-and-boolean-na-handling

PR #224, opened 2018-09-28
Parse all date/time types

max-sixty:
Any date or time value in BQ can be converted to a pandas datetime value. As well as returning dataframes that don't requires as much post-processing, it also reducing the memory footprint of the returned object

I'm not sure whether this will cause any issues around timezones, and whether we need to parse those separately? (though this PR doesn't aim to enable timezone parsing, as long as there's no regression)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 .pytest_cache
 .testmon*
 .vscode/
+.env
 
 # Docs #
 ########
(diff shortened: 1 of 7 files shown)


max-sixty:
This also closes https://github.com/pydata/pandas-gbq/issues/174

max-sixty:
@tswast no rush; ensuring you saw this though

max-sixty:
@tswast gentle ping!

max-sixty:
Merging despite https://github.com/pydata/pandas-gbq/issues/226

Issue #180, opened 2018-05-20
load job config allow quoted newlines 


jwBoral:
https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/load.py

needs to add the condition

job_config.allow_quoted_newlines = True 

current work around is: 
df = df.replace({'\n': ' ', '\r\n': ' ', '\r': ' '}, regex=True)

tswast:
I think we could safely always turn this option on. I see the fact that we serialize to CSV as an implementation detail.

tswast:
Closed by #230

PR #230, opened 2018-10-24
Allow newlines in data passed to to_gbq()

cbandy:
Relates to #180.

I intended to include `\r\n` in the system test, but I found that something was stripping `\r` in the round trip:

```pytest
>       assert result_df.iloc[2,0] == "ij\r\nkl"
E       AssertionError: assert 'ij\nkl' == 'ij\r\nkl'
E         - ij
E         + ij
E         ?   +
E           kl
```

The unit test shows, I think, that we're sending the values correctly, so I didn't investigate further.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+.. _changelog-0.7.1:
+
+0.7.1 / unreleased
+--------------------
+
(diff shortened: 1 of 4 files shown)


cbandy:
@max-sixty fixed in amended commit! thanks.

max-sixty:
Thanks @cbandy (and thanks @tswast for catching, I'm not sure why I missed that notification)

Issue #228, opened 2018-10-19
TST: Migrate to CircleCI for builds


tswast:
Everytime I use Travis, I have the frustration that I can't really define separate Docker images to use as the base for conda versus pip versus different versions of Python. This made changes like https://github.com/pydata/pandas-gbq/pull/222 a lot harder than they would have been if I could have had more control over the base image.

I also suspect that Travis is maybe at fault for our Conda build failing https://github.com/pydata/pandas-gbq/issues/189. My only evidence for this is that I've tested on conda locally several times and am unable to reproduce the issue we see in the Travis build.

There are two interacting features of CircleCI that I think would be a big help:

1. [Docker executor type](https://circleci.com/docs/2.0/executor-types/#using-docker)
2. [Workflows](https://circleci.com/docs/2.0/workflows/), which provide the ability to run multiple "jobs" in parallel, where each job has its own executor definition.

One problem with this plan is that I don't have access to the GCP project that we use for Travis builds so that I can set up the secrets on CircleCI. Maybe @jreback could give me access to that?

tswast:
pandas is mostly migrating to Azure Pipelines: https://github.com/pandas-dev/pandas/issues/22690

max-sixty:
Azure would allow us to test Windows too.

We use Google Cloud Builder successfully internally - but I'm not sure there's a mature OSS offering. Is that right?

tswast:
> Azure would allow us to test Windows too.

Yeah. I'm not too keen on having to maintain VM images, but it'll be okay if we can re-use the configs from pandas.

> We use Google Cloud Builder successfully internally - but I'm not sure there's a mature OSS offering. Is that right?

There's a free tier for Cloud Builder, but nothing OSS-specific AFAIK.

tswast:
I just mailed https://github.com/pydata/pandas-gbq/pull/232. The reason I went with CircleCI instead of Azure is that I couldn't "install" the Azure GitHub app. Even though it's free for OSS projects, I can't install it due to not being a billing admin for the pydata org. CircleCI does not have this requirement. I can enable builds directly from the CircleCI interface without going through GitHub for "billing".

[...]

PR #232, opened 2018-11-02
TST: configure CircleCI

tswast:
Use CircleCI for testing pandas-gbq. This configuration uses the
canonical Python and miniconda3 Docker images to reduce problems caused
by a unique configuration.

Also:

* Re-enables conda tests.
* Adds Python 3.7 tests.

**Edit:** Conda build still fails https://github.com/pydata/pandas-gbq/issues/189

Closes https://github.com/pydata/pandas-gbq/issues/228 https://github.com/pydata/pandas-gbq/issues/197

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,75 @@
+version: 2
+jobs:
+  # Pip
+  "pip-2.7-0.19.2":
+    docker:
+      - image: python:2.7
+    environment:
+      PYTHON: "2.7"
+      PANDAS: "0.19.2"
+    steps:
+      - checkout
+      - run: ci/config_auth.sh
+      - run: ci/run_pip.sh
+  "pip-3.5-0.18.1":
+    docker:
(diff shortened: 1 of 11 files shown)


tswast:
Tests with credentials: https://circleci.com/gh/tswast/pandas-gbq/tree/issue228-circle

max-sixty:
This looks great!! Thanks!

It looks like CircleCI links to builds under @tswast , rather than pydata. That's excellent: it'll test under the committer's credentials (IIRC Travis wouldn't run integration tests on PRs)

It looks like it's not running properly under pydata. I can't seem to change this, I think because I'm a member but not an admin, is that right?

image

tswast:
I think the 1.0 message will go away once this is merged. I see that there was a configuration option to turn on builds (but not share secrets for security) on pull requests. I've just enabled that option, so future PRs should build (without credentials) on CircleCI.

PR #233, opened 2018-11-09
remove old docs

max-sixty:
Pointed out by @msegado https://github.com/pydata/pandas-gbq/issues/174#issuecomment-437252820


max-sixty:
Oh no, @codecov-io is back with its erroneous reports. Anyone have experience in disabling or fixing?

msegado:
Very welcome! And yes, I agree that it was a reasonable decision to match the Pandas behavior.

PR #234, opened 2018-11-09
TST: remove references to Travis. Fix conda tests.

### LINKED ISSUE #228: TST: Migrate to CircleCI for builds

tswast:
Removes the last references to Travis from the README and configuration
files.

I've also attempted to fix the conda build by updating the null floats test to have at least 1 non-null value.

Closes #228


tswast:
Tests running with credentials on my personal CircleCI config: https://circleci.com/workflow-run/921dfde4-498f-4135-99b2-f2306402a375

max-sixty:
Thanks @tswast

Issue #195, opened 2018-08-03
Make Standard SQL Default


RafaelAMello:
Heya!

In the new BigQuery UI Google is rolling out standard SQL is the default for queries, I am wondering if we can do the same with operations using this package?

This is a small connivance but it gets me every time.

Thanks!

tswast:
It is a breaking change so we’ll want to communicate it widely, but I’m supportive of this proposal.

max-sixty:
Shall we:

1. Change default to `None` and warn on `None` that future versions will have Standard SQL default
2.  In a later version change the default to Standard

@RafaelAMello if you agree, would you be interested in creating a PR?

RafaelAMello:
Yup happy to do it,
Should I implement something like the verbose warning depreciation?

When do you think it's best then change the default parameter to `standard`?

Thanks!

max-sixty:
Great, something like `warnings.warn(msg, FutureWarning)` would work.

If @tswast agrees, I think we change the default in 0.6, assuming that's a few months away? 

If someone upgrades past the warning, their code will obviously break rather than return an incorrect result (although won't be clear why it broke) so it's bad but not disastrous.

tswast:
> Change default to None and warn on None that future versions will have Standard SQL default
In a later version change the default to Standard

Sounds like a good plan to me.

max-sixty:
Let's leave this open to track changing the default; the warning is merged, courtesy of @tswast 

(should we do a release with this?)

tswast:
I actually made a release last week. https://pypi.org/project/pandas-gbq/0.6.0/ I forgot to add the release info to GitHub, though. I'll do that now.

max-sixty:
Great, thanks!

tswast:
Now that we have the `pandas_gbq.context` object (#208), we could add a `dialect` property to the context object to allow people to pick their own defaults, as suggested in #103.

To start out, we could populate the default with 'legacy' and switch to 'standard' in a few months.

tswast:
We need to update the samples at https://pandas-gbq.readthedocs.io/en/latest/reading.html to show an explicit dialect.

max-sixty:
Good point @tswast

We could also change the default now?

tswast:
Now that #235 is in and released as part of 0.8 it is possible for people to change the default dialect back to legacy if they really need to. We can now change the default to standard SQL.

PR #235, opened 2018-11-09
ENH: Add pandas_gbq.context.dialect to modify default dialect

tswast:
This will allow people to revert the dialect to legacy SQL when we
switch the default or use standard SQL as the default earlier.

Towards #195.


max-sixty:
Great! 

What do you think about changing the default on master / next release?

tswast:
I wanted something like this change to be in before flipping the default. I kind of want a release with this change in it first before flipping it on master.

PR #236, opened 2018-11-10
Release 0.8.0

tswast:
* Update changelog.
* Add Python 3.7 to supported versions in the package.
* Fix docstrings in new Context class to match pandas-style.
* Document new context class in authentication and API docs.

Note: I put Monday as the date that I'll actually pull the trigger to push to PyPI because it's probably not the kindest thing to release late on a Friday.


max-sixty:
Nice, thanks @tswast !

PR #245, opened 2019-01-25
ENH: Use standard SQL as default.

### LINKED ISSUE #195: Make Standard SQL Default

tswast:
Removes the warning when a SQL dialect is unspecified.

Closes #195


tswast:
CC @RafaelAMello

Issue #242, opened 2019-01-04
Discussion: how to handle the new Int64 (nullable integer) dtype with pandas 0.24.0


tswast:
Currently unreleased, but pandas 0.24.0 will add an extension dtype to allow a nullable integer dtype:  http://pandas-docs.github.io/pandas-docs-travis/integer_na.html#integer-na Unfortunately, we won't use it with our current logic of deferring to the DataFrame constructor for type inference.

> It [Int64, nullable integer] is not the default dtype for integers, and will not be inferred; you must explicitly pass the dtype into array() or Series.

The question is how can we support this dtype in pandas-gbq? I see a few options.

- Use `pd.Int64Dtype()` by default for nullable integer columns, similar to how previously pandas-gbq defaulted to `string` for integer columns.
  - Con: ties new versions of pandas-gbq to 0.24.0+
- Use `pd.Int64Dtype()` for nullable integer columns when pandas-gbq 0.24.0+ is installed.
  - Con: inconsistent with pandas.
  - Con: unable to turn this feature off when float is desired (perhaps for performance reasons).
- Add an argument to `read_gbq` which is a map of column names to dtypes, overriding the dtype of any column present.
  - Con: float isn't the safest default for nullable integer columns, but at least it's consistent with pandas.
  - Con: will require reading rows into separate Series before constructing a DataFrame, as the DataFrame constructor only accepts a single dtype.

tswast:
FWIW: I'm leaning towards "Add an argument to `read_gbq` which is a map of column names to dtypes" because I think that'd be a useful feature in general (perhaps to support other extension dtypes such as for GEOGRAPHY) and also because I'd prefer to stay consistent with pandas in our default behavior.

tswast:
Related to https://github.com/pydata/pandas-gbq/issues/149, does `to_parquet` support the new nullable integer type?

max-sixty:
Great! Agree with your suggestion. We can always choose to default to `pd.Int64Dtype()` in the future, but I would lean towards waiting until pandas uses that type by default

(Sorry to be out of the discussion for a bit)

Issue #133, opened 2018-02-20
Performance 


max-sixty:
We're starting to use BigQuery heavily but becoming increasingly 'bottlenecked' with the performance of moving moderate amounts of data from BigQuery to python.

Here's a few stats:
- **29.1s**: Pulling 500k rows with 3 columns of data (with cached data) using pandas-gbq 
- **36.5s**: Pulling the same query with `google-cloud-bigquery` - i.e. `client.query(query)..to_dataframe()` 
- **2.4s**: Pulling very similar data - same types, same size, from our existing MSSQL box hosted in AWS  (using `pd.read_sql`). That's on standard drivers, nothing like `turbodbc` involved

...so using BigQuery with python is *at least an order of magnitude slower than traditional DBs*.

We've tried exporting tables to CSV on GCS and reading those in, which works fairly well for data processes, though not for exploration.

 A few questions - feel free to jump in with partial replies:
- Are these results expected, or are we doing something very wrong? 
- My prior is that a lot of this slowdown is caused by pulling in HTTP pages, converting to python objects, and then writing those into arrays. Is this approach really scalable? Should `pandas-gbq` invest resources into getting a format that's query-able in exploratory workflows that can deal with more reasonable datasets? (or at least encourage Google to)

tswast:
For some more stats, we can try the running the code in the "benchmark" folder for `google-cloud-python`. https://github.com/GoogleCloudPlatform/google-cloud-python/tree/6391eeef863d0c1b71cb0c6fa86f4d9831b814fa/bigquery/benchmark That should give us a baseline.

maxim-lian:
Need to look at this more, but TensorFlow looks like they've built something that may be fast, in C: https://github.com/tensorflow/tensorflow/commit/1b881b7c77bd1e664382785447a170de2b85f688

tswast:
Duplicate of https://github.com/pydata/pandas-gbq/issues/66?

maxim-lian:
Yes. I've done some work here and can improve time by about 2x and stop the bloating of memory. I'll post something soon-ish. Let's leave this open for the stats above

maxim-lian:
I haven't got around to getting this is a perfect state, but here's a dump for the moment:

### Requirements

In order to go from BigQuery to Pandas we, at a minimum, need to:
- Get some JSON over HTTP
- Decode the JSON into Python objects
- Create an array from the Python objects

### Dos & Don'ts

[...]

PR #247, opened 2019-01-25
CLN: Use `to_dataframe` to download query results.

tswast:
This allows us to remove logic for parsing the schema and align with
google-cloud-bigquery.

Towards #149 

Note, I ran the profiler before and after. For some reason I got about an extra 20 seconds with this versus before (331 before, 352 after). I can't figure out why that would be since in both cases we're just looping over all the rows and constructing a dataframe. Maybe it was just a fluke of my work WiFi network?

In either case, this change will help us with https://github.com/pydata/pandas-gbq/issues/242 once I release the next version of `google-cloud-bigquery`. It'll also help us with https://github.com/pydata/pandas-gbq/issues/133 whenever the mystery faster API launches.

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -35,7 +35,7 @@ def _check_google_client_version():
         raise ImportError("Could not import pkg_resources (setuptools).")
 
     # https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/bigquery/CHANGELOG.md
-    bigquery_minimum_version = pkg_resources.parse_version("0.32.0")
+    bigquery_minimum_version = pkg_resources.parse_version("1.9.0")
     BIGQUERY_INSTALLED_VERSION = pkg_resources.get_distribution(
         "google-cloud-bigquery"
     ).parsed_version
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -679,10 +679,7 @@ def _cast_empty_df_dtypes(schema_fields, df):
             "DataFrame must be empty in order to cast non-nullsafe dtypes"
         )
 
-    dtype_map = {
-        "BOOLEAN": bool,
-        "INTEGER": np.int64,
-    }
+    dtype_map = {"BOOLEAN": bool, "INTEGER": np.int64}
 
     for field in schema_fields:
         column = str(field["name"])
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -482,8 +482,12 @@ class GbqConnector(object):
             self.process_http_error(ex)
 
         schema_fields = [field.to_api_repr() for field in rows_iter.schema]
-        dtypes = _bqschema_to_dtypes(schema_fields)
-        df = rows_iter.to_dataframe(dtypes=dtypes)
+        nullsafe_dtypes = _bqschema_to_nullsafe_dtypes(schema_fields)
+        df = rows_iter.to_dataframe(dtypes=nullsafe_dtypes)
+
+        if df.empty:
+            df = _cast_empty_df_dtypes(schema_fields, df)
+
         logger.debug("Got {} rows.\n".format(rows_iter.total_rows))
         return df
 
@@ -633,11 +637,11 @@ class GbqConnector(object):
         table.create(table_id, table_schema)
 
 
-def _bqschema_to_dtypes(schema_fields):
+def _bqschema_to_nullsafe_dtypes(schema_fields):
     # Only specify dtype when the dtype allows nulls. Otherwise, use pandas's
     # default dtype choice.
     #
-    # see:
+    # See:
     # http://pandas.pydata.org/pandas-docs/dev/missing_data.html
     # #missing-data-casting-rules-and-indexing
     dtype_map = {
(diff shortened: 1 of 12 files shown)


tswast:
There are a couple of errors where the `dtype` that `google-cloud-bigquery` returns is different from the one `pandas-gbq` returns.

The one that most concerns me is the following:

```
>       tm.assert_frame_equal(df, DataFrame({"null_timestamp": [NaT]}))
E       AssertionError: Attributes are different
E       
E       Attribute "dtype" are different
E       [left]:  object
E       [right]: datetime64[ns]
```

`google-cloud-bigquery` returns an object `dtype` for nullable timestamp columns. I think the solution here is to retain `pandas-gbq`'s schema defaults and use the new ability to select a `dtype` in the the `google-cloud-bigquery` `to_dataframe` method.

tswast:
I've updated the tests to match the new dtypes. The main difference is that we now use `datetime64[ns, UTC] for TIMESTAMP columns.

tswast:
I don't know what's going on with CircleCI right now. It can't find `versioneer.py` for some reason.

tests/system/test_gbq.py
```diff
@@ -541,18 +588,13 @@ def test_zero_rows(self, project_id):
...
-                ("ts", "M8[ns]"),
-            ],
-        )
-        expected_result = DataFrame(
-            page_array, columns=["title", "id", "is_bot", "ts"]
-        )
+        empty_columns = {
+            "title": pandas.Series([], dtype=object),
+            "id": pandas.Series([], dtype=object),
+            "is_bot": pandas.Series([], dtype=object),
```
max-sixty: But isn't this a worse result that previous - that an empty column reverts to `object` rather than its original type?

tswast: Yes, it is worse. We'd have to add a second type-conversion step for any non-null object columns. I believe this case only happens on empty dataframes, so I guess it wouldn't be the worst to check for that.

max-sixty: I am definitely keen on doing this well once, rather than us & `google-cloud-bigquery` repeating logic.

But here, it seems like `pandas-gbq` logic is _better_ - to read the schema, and then use that schema in the dataframe. No?

tswast: Done in https://github.com/pydata/pandas-gbq/pull/247/commits/0a6e8d91c3417977880525ea88bc17feba5ebcee

max-sixty: Great! 

Should we do this on every dataframe; no need to have a special case for empty ones?

tswast: > Should we do this on every dataframe; no need to have a special case for empty ones?

We could (if we tell [astype()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.astype.html#pandas.Series.astype) to ignore errors, but I think it's unnecessary. The dataframe constructor seems to pick the right type when bools and ints are passed in: object if it contains nulls, bool/int64 otherwise.

I'm very tempted to use the new nullable integer type by default at https://github.com/pydata/pandas-gbq/issues/242, but probably want to wait for pandas to make that change. For the cases where folks want a very particular dtype, I propose in https://github.com/pydata/pandas-gbq/issues/242 that we add a dtypes argument, much as we did in google-cloud-bigquery.

max-sixty: OK great, thanks @tswast

tswast:
I think the conda failure uncovered a real known issue in https://github.com/pandas-dev/pandas/issues/12513. The workaround is to use the previous behavior of timezone-naive datetimes for BQ TIMEZONE columns, even though that's not exactly correct.

tswast:
@max-sixty Tests are finally green. Mind doing another review before I squash and merge?

max-sixty:
Interesting. I haven't hit that before, but don't use conda, so I guess that explains it. I think the workaround is fine in the meantime - we'll hear about it if anyone strongly needs the tz

pandas_gbq/gbq.py
```diff
@@ -638,45 +637,62 @@ def delete_and_recreate_table(self, dataset_id, table_id, table_schema):
...
+    # default dtype choice.
+    #
+    # See:
     # http://pandas.pydata.org/pandas-docs/dev/missing_data.html
     # #missing-data-casting-rules-and-indexing
     dtype_map = {
         "FLOAT": np.dtype(float),
+        # Even though TIMESTAMPs are timezone-aware in BigQuery, pandas doesn't
+        # support datetime64[ns, UTC] as dtype in DataFrame constructors. See:
+        # https://github.com/pandas-dev/pandas/issues/12513
```
jreback: this is not generally an issue as this is a single dtype for all columns; you are much better off constructing a properly dtypes Series in order to use the tz-aware dtypes

furthermore you need to be really clear if this is a localized timestamp or not ; see the sql conversion code in pandas which handles all of this correctly

tswast: Yeah, I wasn't having an issue except on the conda build with pandas pre-wheels. From https://github.com/pandas-dev/pandas/issues/12513#issuecomment-466515317 you can see in the stacktrace that this is actually constructing `Series` and then combining them as a DataFrame. It's the `Series` constructor that's failing on `dtype='datetime64[ns, UTC]'`.

```
/opt/conda/envs/test-environment/lib/python3.6/site-packages/google/cloud/bigquery/table.py:1325: in _to_dataframe_dtypes
    columns[column] = pandas.Series(columns[column], dtype=dtypes[column])
/opt/conda/envs/test-environment/lib/python3.6/site-packages/pandas/core/series.py:248: in __init__
    raise_cast_failure=True)
/opt/conda/envs/test-environment/lib/python3.6/site-packages/pandas/core/series.py:2967: in _sanitize_array
    subarr = _try_cast(data, False)
```

I had failures on the DataFrame constructor too in the unit tests. If it had worked to construct a series, you're right that I'd prefer to have these be tz-aware.

jreback: you must have an old version
this was worked out a while back

tswast: OK. I'll double-check the Conda CI config.

I kind of prefer to leave this as-is (not tz-aware) as that's the existing behavior. https://github.com/pydata/pandas-gbq/blob/f729a44a48744acc2898350fbfbded791d900967/pandas_gbq/gbq.py#L647 I should make a separate issue for changing that for TIMESTAMP columns.

PR #256, opened 2019-03-12
Remove import sys in logging documentation example

JohnPaton:
The import of `sys` isn't necessary in the logging example, as `sys.stdout` is the default stream for `StreamHandler`. In the interest of keeping examples as minimal as possible I would suggest to remove references to `sys.stdout` and leave that to the `logging` documentation.


max-sixty:
Nice, thanks for the PR @JohnPaton , cleaner is better!

Issue #218, opened 2018-09-17
Allow partial schemas in to_gbq() table_schema


melissachang:
I am using pandas with BigQuery tables. Because pandas doesn't support boolean missing values, BigQuery BOOLEAN columns become dataframe object columns. I have to use to_gbq() table_schema to convert force those columns back to BOOLEAN.

My dataframe has 62 columns. table_schema has to include every column. It would be nice if I could only pass in the boolean columns in table_schema; for the other columns, to_gbq should default to dataframe column type.

tswast:
Currently, if not all columns are specified in the `table_schema`, you get an error, correct? I think it's reasonable to allow specifying a subset and treating them as overrides, leaving the others unchanged.

melissachang:
Yes, if not all columns are specified, you get an error.

For completeness, I want column order to be preserved. Eg say my df has columns A, B, C, D, E. I pass in table schema with C. I want output BigQuery table to have column order A, B, C, D, E.

tswast:
Closed by https://github.com/pydata/pandas-gbq/pull/257

PR #257, opened 2019-03-12
ENH: Allow partial table schema in to_gbq() table_schema (#218)

JohnPaton:
This PR:
* Adds a function to `pandas_gbq.schema`, `update_schema` which takes and old schema and a new one, updating any existing fields and adding any new ones
* Generates the default schema using `schema.generate_bq_schema` for every DataFrame passed to `to_gbq`, and if a `table_schema` is passed, this schema is used to update the generated schema. This way custom provided fields in `table_schema` are added, and missing fields fall back to the generated defaults.

This is the functionality requested in #218

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -20,6 +20,11 @@ Internal changes
 - Use ``to_dataframe()`` from ``google-cloud-bigquery`` in the ``read_gbq()``
   function. (:issue:`247`)
 
+Enhancements
+~~~~~~~~~~~~
+- Allow ``table_schema`` in :func:`to_gbq` to contain only a subset of columns,
+  with the rest being populated using the DataFrame dtypes (:issue:`218`,
+  contributed by @johnpaton)
 
 .. _changelog-0.9.0:
 
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -23,8 +23,8 @@ Internal changes
 Enhancements
 ~~~~~~~~~~~~
 - Allow ``table_schema`` in :func:`to_gbq` to contain only a subset of columns,
-  with the rest being populated using the DataFrame dtypes (:issue:`218`,
-  contributed by @johnpaton)
+  with the rest being populated using the DataFrame dtypes (:issue:`218`) 
+  (contributed by @johnpaton)
 
 .. _changelog-0.9.0:
 
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -1027,7 +1027,9 @@ def to_gbq(
     if not table_schema:
         table_schema = default_schema
     else:
-        table_schema = _update_bq_schema(default_schema, dict(fields=table_schema))
+        table_schema = _update_bq_schema(
+            default_schema, dict(fields=table_schema)
+        )
 
     # If table exists, check if_exists parameter
     if table.exists(table_id):
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -939,9 +939,11 @@ def to_gbq(
         'STRING'},...]``.
         If schema is not provided, it will be
         generated according to dtypes of DataFrame columns.
-        If schema is provided, it must contain all DataFrame columns.
-        pandas_gbq.gbq._generate_bq_schema() may be used to create an initial
-        schema, though it doesn't preserve column order.
+        If schema is provided, it may contain all or a subset of DataFrame
+        columns. If a subset is provided, the rest will be inferred from
+        the DataFrame dtypes.
+        pandas_gbq.gbq._generate_bq_schema() may be used to create an
+        initial schema, though it doesn't preserve column order.
         See BigQuery API documentation on available names of a field.
 
         .. versionadded:: 0.3.1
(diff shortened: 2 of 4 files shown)


pandas_gbq/schema.py
```diff
@@ -31,3 +31,32 @@ def generate_bq_schema(dataframe, default_type="STRING"):
...
+
+    Arguments:
+        schema_old: the old schema to update
+        schema_new: the new schema which will overwrite/extend the old
+    """
+    old_fields = schema_old["fields"]
+    new_fields = schema_new["fields"]
+    output_fields = old_fields.copy()
+
+    field_indices = {field["name"]: i for i, field in enumerate(output_fields)}
```
max-sixty: Do we need `enumerate` here? Are we using a dict as a list?

I don't think it's that important though - the tests look good and so I'm comfortable with the code.

JohnPaton: It's just to get the indices of existing fields by name. Without getting this up front, for every updated field, we have to go through the whole list and look for which one to update. But since `enumerate` gives us the indices, we can just replace the relevant list element directly.

Note that `field_indices` is only being used for updating fields, it's never returned or anything.

pandas_gbq/gbq.py
```diff
@@ -1091,6 +1096,12 @@ def _generate_bq_schema(df, default_type="STRING"):
     return schema.generate_bq_schema(df, default_type=default_type)
 
 
+def _update_bq_schema(schema_old, schema_new):
+    from pandas_gbq import schema
+
+    return schema.update_schema(schema_old, schema_new)
```
max-sixty: Do we need this function? Should we import from `schema` directly? Or there's a circular import?

pandas_gbq/schema.py
max-sixty: 👍

max-sixty:
Great! 

Any thoughts @tswast ?

max-sixty:
@JohnPaton Want to add a Whatsnew? Feel free to add you GH handle

pandas_gbq/gbq.py
JohnPaton: I was just following the pattern used for the only other `schema` function, which is imported the same way in `gbq._generate_bq_schema`. As far as I'm concerned it's fine to get rid of it. Let me know and I'll make the change.

JohnPaton:
@max-sixty By Whatsnew do you just mean to update the Changelog?

JohnPaton:
@max-sixty Done! I didn't see any other handles in the Changelog so if that's not what you meant just let me know 😆

docs/source/changelog.rst
```diff
@@ -237,4 +242,4 @@ Initial release of transfered code from `pandas <https://github.com/pandas-dev/p
 Includes patches since the 0.19.2 release on pandas with the following:
 
 - :func:`read_gbq` now allows query configuration preferences `pandas-GH#14742 <https://github.com/pandas-dev/pandas/pull/14742>`__
-- :func:`read_gbq` now stores ``INTEGER`` columns as ``dtype=object`` if they contain ``NULL`` values. Otherwise they are stored as ``int64``. This prevents precision lost for integers greather than 2**53. Furthermore ``FLOAT`` columns with values above 10**4 are no longer casted to ``int64`` which also caused precision loss `pandas-GH#14064 <https://github.com/pandas-dev/pandas/pull/14064>`__, and `pandas-GH#14305 <https://github.com/pandas-dev/pandas/pull/14305>`__
\ No newline at end of file
+- :func:`read_gbq` now stores ``INTEGER`` columns as ``dtype=object`` if they contain ``NULL`` values. Otherwise they are stored as ``int64``. This prevents precision lost for integers greather than 2**53. Furthermore ``FLOAT`` columns with values above 10**4 are no longer casted to ``int64`` which also caused precision loss `pandas-GH#14064 <https://github.com/pandas-dev/pandas/pull/14064>`__, and `pandas-GH#14305 <https://github.com/pandas-dev/pandas/pull/14305>`__
```
JohnPaton: Not sure why this is showing up as a change, afaik I didn't touch it and I don't see any difference

docs/source/changelog.rst
```diff
@@ -20,6 +20,11 @@ Internal changes
 - Use ``to_dataframe()`` from ``google-cloud-bigquery`` in the ``read_gbq()``
   function. (:issue:`247`)
 
+Enhancements
+~~~~~~~~~~~~
+- Allow ``table_schema`` in :func:`to_gbq` to contain only a subset of columns,
+  with the rest being populated using the DataFrame dtypes (:issue:`218`,
+  contributed by @johnpaton)
```
max-sixty: I _think_ the docs require the issue syntax to have parentheses at the end too, lmk if you know differently though

JohnPaton: No idea... I'll just change it and make a second set of parenthesis

pandas_gbq/gbq.py
max-sixty: I'll defer to Chesterton's fence; we can clean up later if @tswast knows

max-sixty:
Many thanks @JohnPaton !

tswast:
Thanks a bunch for the contribution!

PR #258, opened 2019-03-15
read project_id from credentials if available

daureg:
it seems a bit redundant to have to specify both in the common case where the service account is used for a single project

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -1007,9 +1007,13 @@ def to_gbq(
         raise NotFoundException(
             "Invalid Table Name. Should be of the form 'datasetId.tableId' "
         )
-    if project_id is None and credentials is not None and credentials.__class__.__name__.endswith("Credentials"):
+    has_credentials = (
+        credentials is not None
+        and credentials.__class__.__name__.endswith("Credentials")
+    )
+    if project_id is None and has_credentials:
         project_id = credentials.project_id
-    
+
     connector = GbqConnector(
         project_id,
         reauth=reauth,
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -1007,7 +1007,9 @@ def to_gbq(
         raise NotFoundException(
             "Invalid Table Name. Should be of the form 'datasetId.tableId' "
         )
-
+    if project_id is None and credentials is not None and credentials.__class__.__name__.endswith("Credentials"):
+        project_id = credentials.project_id
+    
     connector = GbqConnector(
         project_id,
         reauth=reauth,
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -318,6 +318,11 @@ class GbqConnector(object):
         self.credentials = credentials
         default_project = None
 
+        # Service account credentials have a project associated with them.
+        # Prefer that project if none was supplied.
+        if self.project_id is None and hasattr(self.credentials, "project_id"):
+            self.project_id = credentials.project_id
+
         # Load credentials from cache.
         if not self.credentials:
             self.credentials = context.credentials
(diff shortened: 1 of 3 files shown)


pandas_gbq/gbq.py
```diff
@@ -1007,6 +1007,12 @@ def to_gbq(
         raise NotFoundException(
             "Invalid Table Name. Should be of the form 'datasetId.tableId' "
         )
+    has_credentials = (
+        credentials is not None
+        and credentials.__class__.__name__.endswith("Credentials")
```
max-sixty: Is there an approach here that is more robust to the name of the class? Do all Credentials classes have a common base we could test for?

tswast: Yeah, they all have https://google-auth.readthedocs.io/en/latest/reference/google.auth.credentials.html#google.auth.credentials.Credentials as a base class.

pandas_gbq/gbq.py
```diff
@@ -1007,6 +1007,12 @@ def to_gbq(
...
+    )
+    if project_id is None and has_credentials:
+        project_id = credentials.project_id
```
tswast: As far as I know, only ServiceAccount credentials actually have a [`project_id`](https://google-auth.readthedocs.io/en/latest/reference/google.oauth2.service_account.html#google.oauth2.service_account.Credentials.project_id) property.

I'd actually much prefer that we do the following condition:

```
if project_id is None and hasattr(credentials, "project_id"):
```

and forget `has_credentials` altogether.

daureg: That's much cleaner indeed

daureg:
I tried to write some test but I'm not sure about them, tell me what you think @tswast

tests/unit/test_gbq.py
```diff
@@ -100,6 +118,45 @@ def test_to_gbq_with_no_project_id_given_should_fail(monkeypatch):
     assert "Could not determine project ID" in str(exception)
 
 
+def test_to_gbq_read_projectid_from_service_account_credentials(
+    min_bq_version
+):
+    import pkg_resources
+
+    credentials, _ = mock_get_service_account_credentials()
+    pandas_version = pkg_resources.parse_version("0.24.0")
```
tswast: Could you remove the version mocking in this and the other test? That was only present in the other tests to look for a specific warning message related to that.

tests/unit/test_gbq.py
```diff
@@ -47,6 +47,24 @@ def mock_get_credentials(*args, **kwargs):
     return mock_credentials, "default-project"
 
 
+def mock_get_service_account_credentials(*args, **kwargs):
```
tswast: Style nit: Let's add `@pytest.fixture` to this and rename to `mock_service_account_credentials`.

It doesn't need to take any arguments, and it only needs to return the `mock_credentials`. Same for `mock_compute_engine_credentials`.

Then, rather than calling these functions in the test, use them as a "pytest fixture" by adding a `mock_service_account_credentials` or `mock_compute_engine_credentials` argument to the test function(s).

tests/unit/test_gbq.py
```diff
@@ -100,6 +118,45 @@ def test_to_gbq_with_no_project_id_given_should_fail(monkeypatch):
...
+        "pkg_resources.Distribution.parsed_version",
+        new_callable=mock.PropertyMock,
+    ) as mock_version:
+        mock_version.side_effect = [min_bq_version, pandas_version]
+        try:
+            gbq.to_gbq(
+                DataFrame([[1]]), "dataset.tablename", credentials=credentials
+            )
+        except gbq.TableCreationError:
+            pass
```
tswast: Another nit. I'd like the test to `assert` something. Could you follow the `test_read_gbq` example instead of `to_gbq` and assert that it returns a DataFrame?

tests/unit/test_gbq.py
tswast: Done in 9035771667204c946d74570ad2498567a0f778f4.

tswast:
Thanks for the contribution @daureg!

PR #259, opened 2019-03-15
DOC: Update "writing" docs with information about schema inference.

tswast:
This commit started as a clean-up change to remove the unnecessary
pandas_gbq.gbq._update_bq_schema method, but I then also updated the
docs for to_gbq to be more clear about how the table_schema argument is
to be used. I added a section to the writing.rst how-to guide about
the table_schema parameter as well.

Some of the "notes" in writing.rst were better as their own subsections.
I moved the note on not to use BigQuery as a transactional database to
the landing page.

I link to the BigQuery sandox docs in the warning about creating
a BigQuery account because you can follow those instructions to use
BigQuery without entering credit card information.


tswast:
Circle is having an outage, but I checked locally that this fixes the system tests.

Issue #261, opened 2019-03-22
TST: dtype for BigQuery TIMESTAMP unexpectedly using datetime64[ns, UTC] dtype


tswast:
```
$ pytest 'tests/system/test_gbq.py::TestReadGBQIntegration::test_return_correct_types[env-current_timestamp()-datetime64[ns]]'
=============================== test session starts ===============================
platform darwin -- Python 3.6.4, pytest-4.2.0, py-1.8.0, pluggy-0.8.1
rootdir: /Users/swast/src/pandas/pandas-gbq, inifile:
collected 1 item                                                                  

tests/system/test_gbq.py F                                                  [100%]

==================================== FAILURES =====================================
_ TestReadGBQIntegration.test_return_correct_types[env-current_timestamp()-datetime64[ns]] _

self = <tests.system.test_gbq.TestReadGBQIntegration object at 0x10c3277b8>
project_id = 'swast-scratch', expression = 'current_timestamp()'
type_ = 'datetime64[ns]'

    @pytest.mark.parametrize(
        "expression, type_",
        [
            ("current_date()", "<M8[ns]"),
            ("current_timestamp()", "datetime64[ns]"),
            ("current_datetime()", "<M8[ns]"),
            ("TRUE", bool),
            ("FALSE", bool),
        ],
    )
    def test_return_correct_types(self, project_id, expression, type_):
        """
        All type checks can be added to this function using additional
        parameters, rather than creating additional functions.
        We can consolidate the existing functions here in time
    
        TODO: time doesn't currently parse
        ("time(12,30,00)", "<M8[ns]"),
        """
        query = "SELECT {} AS _".format(expression)
        df = gbq.read_gbq(
            query,
            project_id=project_id,
            credentials=self.credentials,
            dialect="standard",
        )
>       assert df["_"].dtype == type_
E       AssertionError: assert datetime64[ns, UTC] == 'datetime64[ns]'
E        +  where datetime64[ns, UTC] = 0   2019-03-22 22:35:32.398261+00:00\nName: _, dtype: datetime64[ns, UTC].dtype

tests/system/test_gbq.py:392: AssertionError
============================ 1 failed in 2.68 seconds =============================
```

[...]

PR #263, opened 2019-03-23
ENH: Use tz-aware dtype for timestamp columns

tswast:
I couldn't figure out how *not* to get a tz-aware dtype in 0.24.x
versions, and I wanted a tz-aware dtype anyway for TIMESTAMP, so this
makes it official.

Towards #261.


tswast:
Looks like we'll have to bump the minimum version to `0.20.0`, as `pandas.api.types.DatetimeTZDtype` was first exposed publicly in https://github.com/pandas-dev/pandas/pull/16099

docs/source/reading.rst
```diff
@@ -37,20 +48,39 @@ your job. For more information about query configuration parameters see `here
...
+
+================== =========================
+BigQuery Data Type dtype
+================== =========================
+FLOAT              float
+TIMESTAMP          DatetimeTZDtype(unit='ns', tz='UTC')
+DATETIME           datetime64[ns]
+TIME               datetime64[ns]
+DATE               datetime64[ns]
+================== =========================
```
max-sixty: 👍

tswast:
Oof. Looks like I still get the occasional `TypeError: data type not understood` for all versions before 0.24.0.

@max-sixty Thoughts on bumping the minimum version all the way to 0.24? It's a big jump from what we've had before. Otherwise, I'll see if I can make our tests robust against having naive datetime in 0.23 and before.

tswast:
I'm going to branch this PR with just the docs changes. I'd rather not bump all the way to 0.24.0 just yet.

tswast:
Closing in favor of https://github.com/pydata/pandas-gbq/pull/269. As was pointed out to me in https://github.com/pandas-dev/pandas/issues/25843#issuecomment-476426843 it's more "idiomatic" to call localize rather than be specific about timezones in the dtype. I'd prefer to avoid additional casts / localization if at all possible, so just documenting the current behavior for now.

PR #269, opened 2019-04-03
ENH: Use tz-aware dtype for timestamp columns

tswast:
Adds a table documenting the current behavior, including that pandas 0.24.0 stores as time zone aware dtype and earlier versions store naive. I could not figure out how to make 0.24.0+ store as a naive dtype (https://github.com/pydata/pandas-gbq/issues/261), nor could I figure out how to make earlier versions use time zone aware (https://github.com/pydata/pandas-gbq/pull/263)

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -33,11 +33,14 @@ Internal changes
 
 Enhancements
 ~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


tswast:
Just when I gave up, I figured out how to make it work on old versions. It's not all that bad to call localize on the older versions and this way we don't have to bump the min version.

PR #270, opened 2019-04-04
ENH: Add use_bqstorage_api option to read_gbq

tswast:
The BigQuery Storage API provides a way to read query results quickly
(and using multiple threads). It only works with large query results
(~125 MB), but as of 1.11.1, the google-cloud-bigquery library can
fallback to the BigQuery API to download results when a request to the
BigQuery Storage API fails.

As this API can increase costs (and may not be enabled on the user's
project), this option is disabled by default.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -39,16 +39,12 @@ Enhancements
   (contributed by @johnpaton)
 - Read ``project_id`` in :func:`to_gbq` from provided ``credentials`` if
   available (contributed by @daureg)
-<<<<<<< Updated upstream
 - ``read_gbq`` uses the timezone-aware ``DatetimeTZDtype(unit='ns',
   tz='UTC')`` dtype for BigQuery ``TIMESTAMP`` columns. (:issue:`269`)
-||||||| merged common ancestors
-=======
 - Add ``use_bqstorage_api`` to :func:`read_gbq`. The BigQuery Storage API can
   be used to download large query results (>125 MB) more quickly. If the BQ
   Storage API can't be used, the BigQuery API is used instead. (:issue:`133`,
   :issue:`270`)
->>>>>>> Stashed changes
 
 .. _changelog-0.9.0:
 
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -39,8 +39,16 @@ Enhancements
   (contributed by @johnpaton)
 - Read ``project_id`` in :func:`to_gbq` from provided ``credentials`` if
   available (contributed by @daureg)
+<<<<<<< Updated upstream
 - ``read_gbq`` uses the timezone-aware ``DatetimeTZDtype(unit='ns',
   tz='UTC')`` dtype for BigQuery ``TIMESTAMP`` columns. (:issue:`269`)
+||||||| merged common ancestors
+=======
+- Add ``use_bqstorage_api`` to :func:`read_gbq`. The BigQuery Storage API can
+  be used to download large query results (>125 MB) more quickly. If the BQ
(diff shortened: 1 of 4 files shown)


tswast:
FYI: I ran the test I added with the BQ Storage API on a 4 CPU-node instance.

* `use_bqstorage_api=True` : 76.62 seconds
* `use_bqstorage_api=False` : 265.41 seconds

pandas_gbq/gbq.py
```diff
@@ -815,6 +844,27 @@ def read_gbq(
...
+        You must also have the `bigquery.readsessions.create
+        <https://cloud.google.com/bigquery/docs/access-control#roles>`__
+        permission on the project you are billing queries to.
+
+        **Note:** Due to a `known issue in the ``google-cloud-bigquery``
+        package
+        <https://github.com/googleapis/google-cloud-python/pull/7633>`__
+        (fixed in version 1.11.0), you must write your query results to a
+        destination table. To do this with ``read_gbq``, supply a
+        ``configuration`` dictionary.
```
max-sixty: We could only allow importing this for versions >1.11.0

pandas_gbq/gbq.py
```diff
@@ -727,6 +740,21 @@ def _localize_df(schema_fields, df):
     return df
 
 
+def _make_bqstorage_client(use_bqstorage_api, credentials):
```
max-sixty: Should this take a project too?

docs/source/changelog.rst
```diff
@@ -39,8 +39,16 @@ Enhancements
   (contributed by @johnpaton)
 - Read ``project_id`` in :func:`to_gbq` from provided ``credentials`` if
   available (contributed by @daureg)
+<<<<<<< Updated upstream
```
max-sixty: FYI merge conflict

pandas_gbq/gbq.py
tswast: The BigQuery Storage API client doesn't actually take a project. The project is set when creating a BQ Storage API read session by `to_dataframe` based on the BQ client's project.

Aside: The BQ Storage API client is partially auto-generated, and the client generator doesn't support a default project ID. Since I'm considering the BQ Storage API a lower-level client that's used by the normal BQ client from `to_dataframe`, I felt it wasn't worth adding in that kind of handwritten helper to the autogenned layer.

pandas_gbq/gbq.py
tswast: A few thoughts on this:

* I'd be more comfortable bumping the required version once there's been a few more versions of `google-cloud-bigquery` released past 1.11. I like people to have a few options in case there's a problematic regression in 1.11.1.
* The failure on small results will be quite obvious (`InternalError` stacktrace) if they do try to use with an older version. If someone really wants to try to use it with an older version, they'll find out pretty quickly why they should upgrade, or they really know what they are doing and that they are always going to get results that are large enough for the BigQuery Storage API to read from.
* Someday we might want to make `use_bqstorage_api=True` the default (maybe after the BQ Storage API goes GA?). I think it'd definitely be worth bumping the minimum version at that time.

docs/source/changelog.rst
tswast: Good catch, thanks!

pandas_gbq/gbq.py
max-sixty: Yes that sounds v reasonable. Agree re making the default True when it's GA (or at least soliciting opinions)

tbc I was suggesting _only_ for using bqstorage do we make the min version 1.11 (probably implemented above when attempting the import). Other versions would still run everything else

max-sixty:
Thank you @tswast ! 🚀

PR #278, opened 2019-05-28
Ensure table_schema arg is not modified inplace

### LINKED ISSUE #277: to_gbq should not modify table_schema inplace
Working on a PR...

----

### Problem Description

pandas-gbq version: 0.10.0

GBQ table:

![Screen Shot 2019-05-28 at 2 50 54 PM](https://user-images.githubusercontent.com/25164676/58505950-2fc62580-815c-11e9-9381-f53b36cdd674.png)


Reproducible example:

```python
from copy import deepcopy
import datetime
import pandas_gbq
import pandas as pd
from google.oauth2.service_account import Credentials

df = pd.DataFrame(
    {
        "field1": ["a", "b"],
        "field2": [1, 2],
        "field3": [datetime.date(2019, 1, 1), datetime.date(2019, 5, 1)],
    }
)

original_schema = [
    {
        "name": "field1",
        "type": "STRING",
        "mode": "REQUIRED",
    },
    {
        "name": "field2",
        "type": "INTEGER",
    },
    {
        "name": "field3",
        "type": "DATE",
    },
]
original_schema_cp = deepcopy(original_schema)

pandas_gbq.to_gbq(
    dataframe=df,
    destination_table="XXXXX.schematest",
    project_id="XXXXX",
    credentials=Credentials.from_service_account_file("XXXXX.json"),
    if_exists="append",
    table_schema=original_schema,
)
```

Results:

```python
>>> original_schema                                                                                                                                                                                                                                          
[{'name': 'field1', 'type': 'STRING', 'mode': 'REQUIRED'},
 {'name': 'field2', 'type': 'INTEGER', 'mode': 'NULLABLE'},
 {'name': 'field3', 'type': 'DATE', 'mode': 'NULLABLE'}]
>>> original_schema_cp == original_schema                                                                                                                                                                                                                    
False
```

Ahhhhh!  Nowhere is it noted in https://pandas-gbq.readthedocs.io/en/latest/api.html that, in some situations, the object passed to `table_schema` may be modified in place.  Most libraries go out of their way to avoid modifying mutable arguments; pandas-gbq should do the same.

### Debugging

A `pdb` session shows that `table_schema` is modified in [`connector.load_data()`](https://github.com/pydata/pandas-gbq/blob/59228d9c20cee12b24caa5cc41d3f2e6c0337932/pandas_gbq/gbq.py#L1153), which in turn calls [`load_chunks()`](https://github.com/bsolomon1124/pandas-gbq/blob/59228d9c20cee12b24caa5cc41d3f2e6c0337932/pandas_gbq/gbq.py#L530).  The specific place in `load_chunks()` where the inadvertant modification happens is here:

```python
# https://github.com/bsolomon1124/pandas-gbq/blob/59228d9c20cee12b24caa5cc41d3f2e6c0337932/pandas_gbq/load.py#L72
    for field in schema["fields"]:
        if "mode" not in field:
        field["mode"] = "NULLABLE"
```

This is because, I think, in `table_schema = schema.update_schema(default_schema, dict(fields=table_schema))`, the new `table_schema` now contains a reference to the `dict` elements of the original `table_schema` argument.

bsolomon1124:
Fixes #277.

If any dictionary entry in the `table_schema` arg did
not contain a "mode" key, a mode="NULLABLE" kv pair
would be created; because `schema.update_schema()`
returns an object with references to its mutable input,
this allows the argument to ultimately be modified
by the function rather than the caller.

This pattern was used in both gbq.py and load.py,
so it was refactored into a helper function in
schema.py, which now returns a modified *copy*.
Deepcopy is used because the input is a list of
dictionaries, so a shallow copy would be insufficient.

### DIFF
--- a/pandas_gbq/load.py
+++ b/pandas_gbq/load.py
@@ -66,12 +66,7 @@ def load_chunks(
     if schema is None:
         schema = pandas_gbq.schema.generate_bq_schema(dataframe)
 
-    # Manually create the schema objects, adding NULLABLE mode
-    # as a workaround for
-    # https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4456
-    for field in schema["fields"]:
-        if "mode" not in field:
-            field["mode"] = "NULLABLE"
+    schema = pandas_gbq.schema.add_default_nullable_mode(schema)
 
     job_config.schema = [
         bigquery.SchemaField.from_api_repr(field) for field in schema["fields"]
--- a/tests/unit/test_gbq.py
+++ b/tests/unit/test_gbq.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import copy
+import datetime
 from unittest import mock
 
 import numpy
@@ -477,3 +479,45 @@ def test_generate_bq_schema_deprecated():
     with pytest.warns(FutureWarning):
         df = DataFrame([[1, "two"], [3, "four"]])
         gbq.generate_bq_schema(df)
+
+
+def test_load_does_not_modify_schema_arg():
+    # Test of Issue # 277
+    df = DataFrame(
+        {
+            "field1": ["a", "b"],
+            "field2": [1, 2],
+            "field3": [datetime.date(2019, 1, 1), datetime.date(2019, 5, 1)],
+        }
+    )
+    original_schema = [
+        {"name": "field1", "type": "STRING", "mode": "REQUIRED"},
+        {"name": "field2", "type": "INTEGER"},
+        {"name": "field3", "type": "DATE"},
+    ]
+    original_schema_cp = copy.deepcopy(original_schema)
+    gbq.to_gbq(
+        df,
+        "dataset.schematest",
+        project_id="my-project",
+        table_schema=original_schema,
+        if_exists="fail",
+    )
+    assert original_schema == original_schema_cp
+
+    # Test again now that table exists - behavior will differ internally
+    # branch at if table.exists(table_id)
+    original_schema = [
+        {"name": "field1", "type": "STRING", "mode": "REQUIRED"},
+        {"name": "field2", "type": "INTEGER"},
+        {"name": "field3", "type": "DATE"},
+    ]
+    original_schema_cp = copy.deepcopy(original_schema)
+    gbq.to_gbq(
+        df,
+        "dataset.schematest",
+        project_id="my-project",
+        table_schema=original_schema,
+        if_exists="append",
+    )
+    assert original_schema == original_schema_cp
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -13,6 +13,7 @@ except ImportError:  # pragma: NO COVER
     bigquery_storage_v1beta1 = None
 
 from pandas_gbq.exceptions import AccessDenied
+import pandas_gbq.schema
 
 logger = logging.getLogger(__name__)
 
@@ -1269,12 +1270,7 @@ class _Table(GbqConnector):
         table_ref = self.client.dataset(self.dataset_id).table(table_id)
         table = Table(table_ref)
 
-        # Manually create the schema objects, adding NULLABLE mode
-        # as a workaround for
-        # https://github.com/GoogleCloudPlatform/google-cloud-python/issues/4456
-        for field in schema["fields"]:
-            if "mode" not in field:
-                field["mode"] = "NULLABLE"
+        schema = pandas_gbq.schema.add_default_nullable_mode(schema)
 
         table.schema = [
             SchemaField.from_api_repr(field) for field in schema["fields"]
--- a/pandas_gbq/schema.py
+++ b/pandas_gbq/schema.py
@@ -1,5 +1,7 @@
 """Helper methods for BigQuery schemas"""
 
+import copy
+
 
 def generate_bq_schema(dataframe, default_type="STRING"):
     """Given a passed dataframe, generate the associated Google BigQuery schema.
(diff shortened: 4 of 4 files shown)


tests/unit/test_gbq.py
```diff
@@ -477,3 +479,29 @@ def test_generate_bq_schema_deprecated():
...
+    original_schema_cp = copy.deepcopy(original_schema)
+    gbq.to_gbq(
+        df,
+        "dataset.schematest",
+        project_id="my-project",
+        table_schema=original_schema,
+    )
+    assert [sorted(i.items()) for i in original_schema] == [
+        sorted(i.items()) for i in original_schema_cp
+    ]
```
max-sixty: Can we `assert original_schema == original_schema_cp`?

bsolomon1124: Yes, I think that should be okay, but for some reason the original commit/patch isn't doing what it's supposed to and I'm not really sure why.

Here is stepping into the branch locally:

```python
In [3]: original_schema                                                                                                                                      
Out[3]: 
[{'name': 'field1', 'type': 'STRING', 'mode': 'REQUIRED'},
 {'name': 'field2', 'type': 'INTEGER', 'mode': 'NULLABLE'},
 {'name': 'field3', 'type': 'DATE', 'mode': 'NULLABLE'}]
```

So it is still being modified inadvertently despite the `deepcopy()` call.

pandas_gbq/load.py
```diff
@@ -65,6 +66,10 @@ def load_chunks(
 
     if schema is None:
         schema = pandas_gbq.schema.generate_bq_schema(dataframe)
+    else:
+        # See Issue #277
+        # Prevent inplace modification of toplevel function argument
+        schema = copy.deepcopy(schema)
```
tswast: Might make sense to do this at the `to_gbq` level instead, since that's the main public entrypoint.

bsolomon1124: I see what you're saying, but I would argue at the same time you'd want to cure things at the root.  If you have some other library that's calling `load_chunks()` (I guess it's technically public API?) then it may be subject to the same problem even if fixed in `to_gbq()`.

bsolomon1124:
So, I might be thinking too hard about this one @tswast / @max-sixty , but this seems to *only* occur if the GBQ table already exists and the write is done in append-mode.  Giving myself a headache after a long day at work..

Called when `dataset.schematest` does not exist and will be created; in this first case, the argument gets modified:

```python
In [17]: original_schema = [ 
    ...:     {"name": "field1", "type": "STRING", "mode": "REQUIRED"}, 
    ...:     {"name": "field2", "type": "INTEGER"}, 
    ...:     {"name": "field3", "type": "DATE"}, 
    ...: ] 
    ...: gbq.to_gbq(df, "dataset.schematest", project_id="xxx", table_schema=original_schema, if_exists="append")                                  

In [18]: original_schema                                                                                                                                     
Out[18]: 
[{'name': 'field1', 'type': 'STRING', 'mode': 'REQUIRED'},
 {'name': 'field2', 'type': 'INTEGER', 'mode': 'NULLABLE'},
 {'name': 'field3', 'type': 'DATE', 'mode': 'NULLABLE'}]
```

Now, called again directly after the above; seemingly no problem here:

```python
In [19]: df = DataFrame( 
    ...:     { 
    ...:         "field1": ["a", "b"], 
    ...:         "field2": [1, 2], 
    ...:         "field3": [datetime.date(2019, 1, 1), datetime.date(2019, 5, 1)], 
    ...:     } 
    ...: ) 
    ...: original_schema = [ 
    ...:     {"name": "field1", "type": "STRING", "mode": "REQUIRED"}, 
    ...:     {"name": "field2", "type": "INTEGER"}, 
    ...:     {"name": "field3", "type": "DATE"}, 
    ...: ] 
    ...: gbq.to_gbq(df, "dataset.schematest", project_id="xxx", table_schema=original_schema, if_exists="append")                                  

In [20]: original_schema                                                                                                                                     
Out[20]: 
[{'name': 'field1', 'type': 'STRING', 'mode': 'REQUIRED'},
 {'name': 'field2', 'type': 'INTEGER'},
 {'name': 'field3', 'type': 'DATE'}]
```

max-sixty:
I think _no_ function in this library should modify its arguments. 

So we should do the `deepcopy` in the function that does modify them - can we track that down and apply it to that function? I think that's the way of applying @bsolomon1124 's good principle of fixing the problem at its root.

bsolomon1124:
Aha.  It looks like the same pattern is repeated in https://github.com/bsolomon1124/pandas-gbq/blob/488f6b7e984757c79bbc4f59b1121fcd6e3afbbf/pandas_gbq/gbq.py#L1272.  (Seeing some similar patterns in `gbq.py` but not sure if they are doing the same thing; see `grep -nr "for field in schema" .`

tests/unit/test_gbq.py
bsolomon1124: On second check - yes, `assert original_schema == original_schema_cp`.  I was worried about un-orderedness of dictionaries in earlier Python3.  [However](https://docs.python.org/3.4/library/stdtypes.html#dict), 

> Dictionaries compare equal if and only if they have the same (key, value) pairs.

bsolomon1124:
Ok, I think we've got it @max-sixty @tswast .  See the note above regarding `gbq.py` and the updated commit msg, which I've squashed into one.  I do wonder if there are other more subtle cases like this hidden in `gbq.py.  Also, I think that `update_schema()` may be returning data that references `output_fields`, but need to inspect that further.

max-sixty:
That's perfect! Ideal design and implementation. 

I'll merge tomorrow unless anyone has any thoughts.

Re the `output_fields` - I agree, I think it would be better if that returned a new object without any objects in `schema_old`. I think it's less likely to hit than the current issue, but still worth adding a `copy` there. If you want to add a `TODO` / issue, please do; but don't feel we have to wait to fix to merge this.

PR #285, opened 2019-07-26
BUG: Use object dtype for STRING, ARRAY, and STRUCT columns when there are zero rows.

tswast:
If a there are no rows, the default dtype is used (which is now
float64, must previously have been object).


max-sixty:
Nice, thanks @tswast !

PR #291, opened 2019-10-05
Missing parenthesis in code example

camfrout:
The closing parenthesis is missing in the current code example, making copy/paste of the code failing.


max-sixty:
Good spot, thanks @camfrout !

PR #292, opened 2019-10-25
Show progress bar when downloading data

### LINKED ISSUE #182: ENH: show progress while fetching rows for query
As reported by @QuinRiva in https://github.com/pydata/pandas-gbq/issues/12#issuecomment-391212920

Progress is written to logging while a query is running, but no progress is reported between when a query finishes and while the data is being downloaded to be added to a DataFrame. The problem is that we call `list(rows_iter)` to fetch all pages. Previously, progress was written as each page was downloaded.

https://github.com/pydata/pandas-gbq/blob/08166685d3305a57fbfd3bc4c41a1cf5df98ebcf/pandas_gbq/gbq.py#L294-L299

A possible solution is to loop over [`rows_iter.pages`](https://google-cloud-python.readthedocs.io/en/latest/bigquery/reference.html#google.cloud.bigquery.table.RowIterator.pages) instead. After the first page is fetched, the `rows_iter.total_rows` property is available, so it would be possible to display a percent complete or even use `tqdm` as done in https://github.com/pydata/pandas-gbq/pull/166.

dakl:
@tswast Perhaps like so? 

Fixes #182.

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -436,7 +436,7 @@ class GbqConnector(object):
 
         raise GenericGBQException("Reason: {0}".format(ex))
 
-    def run_query(self, query, max_results=None, progress_bar=None, **kwargs):
+    def run_query(self, query, max_results=None, progress_bar_type=None, **kwargs):
         from concurrent.futures import TimeoutError
         from google.auth.exceptions import RefreshError
         from google.cloud import bigquery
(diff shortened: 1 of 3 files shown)


pandas_gbq/gbq.py
```diff
@@ -833,6 +839,7 @@ def read_gbq(
     max_results=None,
     verbose=None,
     private_key=None,
+    progress_bar=None,
```
tswast: Let's call the argument `progress_bar_type` for consistency with google-cloud-bigquery.

Also, I believe it's safe to set the default to `tqdm` even if the library is not installed.

dakl: 👍Good points, done and done.

PR #295, opened 2019-11-23
FIX: close BigQuery Storage client transport channel after use

### LINKED ISSUE #294: BUG: potential resource leak when using BigQuery Storage API
It came to my attention in https://github.com/googleapis/google-cloud-python/issues/9457 that when a gRPC Google Cloud client object is destroyed, it doesn't clean up all the underlying resources from the gRPC transport layer.

I'd like to address this at the Google Cloud client level by providing better support for short-lived gRPC clients, but in the meantime, we need to call `bqstorage_client.transport.close()` when we are done using a BigQuery Storage API client.

tswast:
This fixes a file descriptor leak.

Closes #294

### DIFF
--- a/ci/requirements-3.5.pip
+++ b/ci/requirements-3.5.pip
@@ -1,5 +1,5 @@
 pandas==0.19.0
 google-auth==1.4.1
 google-auth-oauthlib==0.0.1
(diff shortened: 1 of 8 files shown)


max-sixty:
LGTM!

PR #314, opened 2020-02-24
ENH: restrict to Python 3.5 and higher in setup.py

### LINKED ISSUE #306: ENH: restrict to Python 3.5 and higher in setup.py
Some Foreword: please ask me for further information/MWE if necessary, I hope it will be understandable without for now:

We expercienced some error like the following:
```
MYENV/local/lib/python2.7/site-packages/pandas/io/common.pyc in writerows(self, rows)
    549             data = self.encoder.encode(data)
    550             # write to the target stream
--> 551             self.stream.write(data)
    552             # empty queue
    553             self.queue.truncate(0)

TypeError: unicode argument expected, got 'str'
```

This seems to appear in version 0.13.0, pinning back to 0.12.0 seems to work.

Still, referring to the changelog they are both not supported under python 2 (under which this occurred). Is it expected behaviour they get still installed? Is this a problem on pypi side or can it be fixed on the package site?

Thanks for your efforts!

nicoa:
closes #306

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,7 @@ setup(
     keywords="data",
     install_requires=INSTALL_REQUIRES,
     extras_require=extras,
+    python_requires=">=3.5",
     packages=find_packages(exclude=["contrib", "docs", "tests*"]),
     test_suite="tests",
 )


tswast:
Thanks for the contribution!

PR #318, opened 2020-04-25
BUG: use original schema when appending

### LINKED ISSUE #315: Pandas should get the schema from bigquery if pushing to a table that already exists
Right now, when pushing new data to an **already existing table** using `to_gbq`, with option `if_exists=append`, but no explicit `table_schema`, pandas generates a default table schema, where the `mode`  of the column, which takes value either `REQUIRED` or `NULLABLE`, by default is always `NULLABLE`.  

It would make sense for pandas to fetch schema, and apply those for case where `if_exists=append` instead of passing a `NULLABLE` mode.

ShantanuKumar:
Fixes #315 
Don't overwrite table schema when appending to an existing table

TODO:
- [x] Adds relevant tests.
- [x] Lint passes.
- [x] Add to changelog.


ShantanuKumar:
If I lint using `black` locally, it's removing the `u` attached to a unicode string here in the test file
https://github.com/pydata/pandas-gbq/blob/5538469fe147f8b25b89926fdb748a29cfc851dc/tests/system/test_gbq.py#L1106

I had to manually add it back. But it seems the lint checks are failing on `circle-ci` because black is removing it when doing the test on `circle-ci`

tswast:
This is making a few more table GET HTTP requests than necessary. If you don't mind, I'll push a commit or two to this to refactor.

tswast:
I just pushed d8c11aab3699b528bceafdb119ac59936d1449d2 with the suggested refactoring

PR #336, opened 2020-11-06
BUG: use greater precision when serializing floating points

### LINKED ISSUE #326: Floats can lose precision when loading to BigQuery
The float precision is set here: https://github.com/pydata/pandas-gbq/blob/d251db03b159447331ac9ae63e13d295d75bad70/pandas_gbq/load.py#L22

This is insufficient to represent all 64 bit floats without losing precision. For example 26/59 should be represented as `0.4406779661016949` but under this it is represented as `0.440677966101695`.

This was added intentionally [here](https://github.com/pydata/pandas-gbq/pull/193) to fix a different issue but it causes us some issues as we need perfect reconciliation between systems. It seems like it _should_ be possible to get the best of both worlds and output the correct number of digits in all cases.

The original suggestion was to use `%g` but this was changed to `%.15g` – it's not clear to me what the rationale is for that, it seems like `%g` is strictly better but I'm sure I'm missing something.

tswast:
This allows the exact binary representation to be transferred
correctly, round-trip.

- [x] closes #xxxx
- [x] tests added / passed

Closes #326


max-sixty:
Looks great! Thanks!

PR #340, opened 2020-11-09
BUG: support INT64 and other standard SQL aliases in to_gbq table_schema

### LINKED ISSUE #322: df.to_gbq(): InvalidSchema when using 'type': 'INT64' and if_exists = 'append'
Hello there,

I'm facing an issue with the function `df.to_gbq()`

From my understanding, it seems that when we are using `if_exists = 'append'` when the destination_table already exists, we are retrieving the original_schema from destination_table.

The original_schema will contain every fields but type retrieve is `INTEGER` and not `INT64` anymore.
Since the code is simply checking that all fields in the `schema_local` are the same in the `schema_remote`, this conducts to a missmatch and function `schema_is_subset` is returning `False` instead of `True`.

I'm not sure about where we should fix this error, but it's kind of annoying when working with integers values.

tswast:
- [x] closes #xxxx
- [x] tests added / passed

Closes #322


max-sixty:
Looks perfect! Thanks!

PR #347, opened 2020-11-29
ENH: add project id to destination table in to_gbq()

### LINKED ISSUE #321: ENH: Function to_gbq(): Dissociate the project ID to run the BQ job from the table project
I think it would be very helpful in certain situations to differentiate between the project where the BQ job is to be run, and the project in which the the table is located (especially because of the costs billing). 
As far as I understood, at present this is not possible when calling the function `to_gbq` ([link](https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/gbq.py#L998)), because it gets only one `project_id` parameter as input, and the `destination_table` is in the form `dataset.tablename`.
For instance, in the `python-bigquery` API from Google, you can do this by using the method `load_table_from_dataframe` ([link](https://github.com/googleapis/python-bigquery/blob/master/google/cloud/bigquery/client.py#L1780)) which receives as input a `TableReference` object ([link](https://github.com/googleapis/python-bigquery/blob/master/google/cloud/bigquery/table.py#L119)) (with a `project_id`, a `dataset_id` and a `table_id`), being however called from a `client` running jobs in a separate `project_id`.
I think this enhancement would be very useful in situations where splitting the costs among projects is important, while maintaining the availability of the data among projects and/or departments (which is the case in certain organizations).
Many thanks in advance for taking into consideration this issue, and please excuse me if I misunderstood something about the functionality of this method.
Best regards.

vreyespue:
After merging this PR, it will be possible to write data into tables stored in a different project ID than the client project when calling the function `to_gbq()`.
Note that in the python-bigquery API from Google you can do this by using the method `load_table_from_dataframe` ([link](https://github.com/googleapis/python-bigquery/blob/master/google/cloud/bigquery/client.py#L1780)) which receives as input a `TableReference` object ([link](https://github.com/googleapis/python-bigquery/blob/master/google/cloud/bigquery/table.py#L119)) (with a `project_id`, a `dataset_id` and a `table_id`), being however called from a `client` running jobs in a separate `project_id`.
I think this enhancement will be very useful in situations where splitting the costs among projects is important, while maintaining the availability of the data among projects and/or departments (which is the case in some organizations).

- [x] closes #321
- [x] tests added / passed

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -1147,7 +1147,7 @@ def to_gbq(
     bqclient = connector.client
 
     destination_table_ref = bigquery.table.TableReference.from_string(
-        destination_table, default_project=project_id
+        destination_table, default_project=connector.project_id
     )
 
     project_id_table = destination_table_ref.project
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -604,6 +604,7 @@ class GbqConnector(object):
     def load_data(
         self,
         dataframe,
+        project_id_table,
         dataset_id,
         table_id,
         chunksize=None,
@@ -618,6 +619,7 @@ class GbqConnector(object):
             chunks = load.load_chunks(
                 self.client,
                 dataframe,
+                project_id_table,
                 dataset_id,
                 table_id,
                 chunksize=chunksize,
@@ -1037,7 +1039,8 @@ def to_gbq(
     dataframe : pandas.DataFrame
         DataFrame to be written to a Google BigQuery table.
     destination_table : str
-        Name of table to be written, in the form ``dataset.tablename``.
+        Name of table to be written, in the form ``dataset.tablename`` or
+        ``project.dataset.tablename``.
     project_id : str, optional
         Google BigQuery Account project ID. Optional when available from
         the environment.
@@ -1133,7 +1136,8 @@ def to_gbq(
 
     if "." not in destination_table:
         raise NotFoundException(
-            "Invalid Table Name. Should be of the form 'datasetId.tableId' "
+            "Invalid Table Name. Should be of the form 'datasetId.tableId' or "
+            "'projectId.datasetId.tableId'"
         )
 
     connector = GbqConnector(
@@ -1145,7 +1149,14 @@ def to_gbq(
         private_key=private_key,
     )
     bqclient = connector.client
-    dataset_id, table_id = destination_table.rsplit(".", 1)
+
+    destination_table_ref = bigquery.table.TableReference.from_string(
+        destination_table, default_project=project_id
+    )
+
+    project_id_table = destination_table_ref.project
+    dataset_id = destination_table_ref.dataset_id
+    table_id = destination_table_ref.table_id
 
     default_schema = _generate_bq_schema(dataframe)
     if not table_schema:
@@ -1160,7 +1171,7 @@ def to_gbq(
         table = bqclient.get_table(destination_table)
     except google_exceptions.NotFound:
         table_connector = _Table(
-            project_id,
+            project_id_table,
             dataset_id,
             location=location,
             credentials=connector.credentials,
@@ -1203,6 +1214,7 @@ def to_gbq(
 
     connector.load_data(
         dataframe,
+        project_id_table,
         dataset_id,
         table_id,
         chunksize=chunksize,
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -604,7 +604,7 @@ class GbqConnector(object):
     def load_data(
         self,
         dataframe,
-        destination_table,
+        destination_table_ref,
         chunksize=None,
         schema=None,
         progress_bar=True,
@@ -617,9 +617,7 @@ class GbqConnector(object):
             chunks = load.load_chunks(
                 self.client,
                 dataframe,
-                project_id_table,
-                dataset_id,
-                table_id,
+                destination_table_ref,
                 chunksize=chunksize,
                 schema=schema,
                 location=self.location,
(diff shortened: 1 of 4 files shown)


pandas_gbq/load.py
```diff
@@ -50,13 +50,16 @@ def encode_chunks(dataframe, chunksize=None):
...
+    project_id_table,
     dataset_id,
     table_id,
     chunksize=None,
     schema=None,
     location=None,
 ):
-    destination_table = client.dataset(dataset_id).table(table_id)
+    destination_table = (
+        client(project=project_id_table).dataset(dataset_id).table(table_id)
```
max-sixty, line 61: Tests pass so presumably this must work — but are we sure `client` is callable here?

vreyespue: You are totally right, `client` isn't callable here. When trying to use the code in a real example, I get the following error:
```
  File "to_gbq_simple.py", line 42, in <module>
    main(args.project_id, args.table_id)
  File "to_gbq_simple.py", line 33, in main
    pandas_gbq.to_gbq(df, table_id, project_id=project_id)
  File "/Users/vicente.reyes/hub/vreyespue/pandas-gbq/pandas_gbq/gbq.py", line 1222, in to_gbq
    progress_bar=progress_bar,
  File "/Users/vicente.reyes/hub/vreyespue/pandas-gbq/pandas_gbq/gbq.py", line 631, in load_data
    for remaining_rows in chunks:
  File "/Users/vicente.reyes/.pyenv/versions/3.7.5/lib/python3.7/site-packages/tqdm/std.py", line 1104, in __iter__
    for obj in iterable:
  File "/Users/vicente.reyes/hub/vreyespue/pandas-gbq/pandas_gbq/load.py", line 61, in load_chunks
    client(project=project_id_table).dataset(dataset_id).table(table_id)
TypeError: 'Client' object is not callable
```
I will try to find another way of implementing this.
Many thanks and best regards.

vreyespue: Dear @max-sixty, I resolved the issue by using the following statement.
```
destination_table = client.dataset(dataset_id, project_id_table).table(
        table_id
    )
```
I also tested it in a real example, now it is working properly.
Many thanks and best regards.

pandas_gbq/gbq.py
```diff
@@ -604,6 +604,7 @@ def _download_results(
     def load_data(
         self,
         dataframe,
+        project_id_table,
```
max-sixty, line 607: I don't have that strong a view, but I'm not sure if using `project_id_table` is clearer or than just `project_id`

max-sixty:
@vreyespue this looks good, thanks for this. Would you like to add a note to the changelog?

If @tswast gets a chance to review then I'll wait for his view.

pandas_gbq/gbq.py
vreyespue: Many thanks for this sensible comment. I thought that this variable name would make things clearer, because we refer here to the project in which the destination table is located, and not the one associated to the BQ client. However, if you prefer to use another variable name, I would be open for that. Maybe @tswast has an opinion on this? Best regards.

vreyespue:
> @vreyespue this looks good, thanks for this. Would you like to add a note to the changelog?
> 
> If @tswast gets a chance to review then I'll wait for his view.

I agree, it would be great if @tswast can take a look at this. Many thanks and best regards.

pandas_gbq/gbq.py
```diff
@@ -604,6 +604,7 @@ def _download_results(
...
         dataset_id,
         table_id,
```
tswast, line 609: I'd prefer we actually change this to pass in the table reference we create in `to_gbq`

```suggestion
        destination_table,
```

pandas_gbq/load.py
```diff
@@ -50,13 +50,16 @@ def encode_chunks(dataframe, chunksize=None):
 def load_chunks(
     client,
     dataframe,
+    project_id_table,
     dataset_id,
     table_id,
```
tswast, line 55:
```suggestion
    destination_table,
```

pandas_gbq/load.py
```diff
@@ -50,13 +50,16 @@ def encode_chunks(dataframe, chunksize=None):
...
     chunksize=None,
     schema=None,
     location=None,
 ):
-    destination_table = client.dataset(dataset_id).table(table_id)
+    destination_table = client.dataset(dataset_id, project_id_table).table(
+        table_id
+    )
```
tswast, line 62:
```suggestion
```

pandas_gbq/gbq.py
```diff
@@ -1203,6 +1214,7 @@ def to_gbq(
 
     connector.load_data(
         dataframe,
+        project_id_table,
         dataset_id,
         table_id,
```
tswast, line 1219:
```suggestion
        destination_table_ref,
```

tswast (changes requested):

pandas_gbq/load.py
tswast: > Tests pass so presumably this must work — but are we sure client is callable here?

Unfortunately the secrets needed to run the system tests aren't available until merging, so many tests were skipped.

vreyespue:
Hi @tswast, sorry I just applied your suggestions without properly testing. I will fix this tomorrow. I mark the PR as WIP accordingly. Many thanks and best regards.

vreyespue:
Hi @tswast, I resolved the issue 👍 the call to `load.load_chunks` was broken. Now the review can continue. Many thanks and best regards.

vreyespue:
@tswast @max-sixty is there any progress on this issue? I already applied all the requested changes, so I assume this PR is good to be merged, right? Many thanks and best regards.

tswast (approved):
Thanks for the contribution! I've verified locally that the system tests are passing.

I pushed a small commit to check for `project_table` in the added unit test and modifie `_Table` so that the unit test passes (it was still using the client project)

vreyespue:
Ok very cool 😄  I am very happy to help 👍  many thanks and best regards

PR #350, opened 2020-12-22
ENH: show schemas difference when throwing InvalidSchema exception

### LINKED ISSUE #349: Improve InvalidSchema exception by adding specific fields that do not match
I am talking about this [InvalidSchema](https://github.com/pydata/pandas-gbq/blob/master/pandas_gbq/gbq.py#L1194) exception.

It may be useful to show at most 3 (for example) fields that have different types. It should help to debug a type mismatch faster.

artemrys:
After merging this PR - `to_gbq` method will show a more detailed description of an error in case of `append` mode.

In particular, it will show at most 3 differences in schemas (both missing field and the same field but a different type) and will indicate how many more differences left (if any).

In #349 there was a suggestion to use `set` to compare schemas, but the elements have `dict` type, so they are not hashable. So I decided to go with a more straightforward solution and compare dataframe and BigQuery fields one by one. And I am doing this only if the dataframe schema is not a subset of a BigQuery schema to leave the number of successful path operations the same.

Do you want me to add more system-level tests to verify different exception messages or is it enough to have them covered by unit tests? 

- [x] closes #349
- [x] tests added / passed


pandas_gbq/schema.py
```diff
@@ -42,8 +49,99 @@ def _clean_schema_fields(fields):
...
+            schema_difference.append("And {} more left.".format(diffs_left))
+        diff_to_show = "\n".join(schema_difference)
+    return diff_to_show
+
+
+def _calculate_schema_difference(schema_remote, schema_local):
+    """Calculates difference in dataframe and BigQuery schemas.
+
+    Compares dataframe and BigQuery schemas to identify exact differences
+    in each field (field can be missing in the dataframe or field can have
```
tswast, line 106: Missing fields in the dataframe should be OK, right? It's extra fields in the dataframe that can be a problem (unless we allow field addition, as requested here: https://github.com/pydata/pandas-gbq/issues/107)

pandas_gbq/schema.py
```diff
@@ -42,8 +49,99 @@ def _clean_schema_fields(fields):
...
+    List[str]
+        List of field differences
+    """
+    fields_remote = _clean_schema_fields(schema_remote.get("fields", []))
+    fields_local = _clean_schema_fields(schema_local.get("fields", []))
+    diff = []
+    for field_remote in fields_remote:
+        for field_local in fields_local:
+            if field_local["name"] == field_remote["name"]:
+                if field_local["type"] != field_remote["type"]:
```
tswast, line 129: This might miss type mismatches if the order of the columns is different in either `field_remote` or `field_local` (or does `_clean_schema_fields` sort them?). I think we'd want:

1. Loop through only `fields_local` (dataframe)
2. Check if the field name isn't in `fields_remote` (table)
3. Check if the field types don't match

tests/unit/test_schema.py
```diff
@@ -69,6 +69,79 @@ def test_schema_is_subset_fails_if_not_subset(module_under_test):
...
+                {"name": "B", "type": "FLOAT"},
+            ],
+            (
+                "Field 'B' has different types: dataframe 'FLOAT', BigQuery 'STRING'.\n"
+                "Field 'C': no such field in the dataframe."
+            ),
+        ),
+        (
+            [
+                {"name": "A", "type": "FLOAT"},
```
tswast, line 115: I think we'll want some tests where the order of the columns doesn't line up. That way we can be sure we aren't showing errors that aren't actually errors.

tswast (approved):
Thanks for the contribution! I left some feedback in the review. I'm concerned about the case where column orders don't line up.

tswast (changes requested):
Oops. Last review was meant to be request changes

meredithslota:
Checking in - this PR has set for some time. Close? Revise?

artemrys:
Hey, sorry, unfortunately, I do not have time to work on this one

PR #380, opened 2021-08-26
chore: enable owlbot and Kokoro

tswast:
Towards googleapis/python-bigquery-pandas#367

PR #393, opened 2021-09-28
test: use templated noxfiles

tswast:
Adjust test configuration to prepare for running on Kokoro.

Moves manual changes (+ noxfiles) out of #380


tswast:
Note: Kokoro failures are expected, as there are no config files for Kokoro yet. Circle CI should still be functioning, though.

samples/snippets/noxfile.py
```diff
@@ -0,0 +1,266 @@
+# Copyright 2019 Google LLC
```
tswast, line 1: This file is templated: https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_samples/noxfile.py.j2

noxfile.py
```diff
@@ -1,14 +1,24 @@
-# Copyright (c) 2017 pandas-gbq Authors All rights reserved.
-# Use of this source code is governed by a BSD-style
-# license that can be found in the LICENSE file.
-
-"""Nox test automation configuration.
-
-See: https://nox.readthedocs.io/en/latest/
-"""
-
+# -*- coding: utf-8 -*-
```
tswast, line 1: This file is templated: https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/noxfile.py.j2

tswast:
Other than the two noxfiles, these changes are manual.

.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     steps:
       - checkout
       - run: ci/config_auth.sh
-      - run: nox -s unit-3.8 system-3.8 cover
+      - run: PYTHON=3.8 ci/run_pip.sh
```
tseaver, line 42: What is the rationale for avoiding `nox` under CI?

tswast: Good question. I forget why I did this. I'll switch back to `nox` and see if anything breaks.

tswast: It was because `run_pip.sh` calls `run_tests.sh`, which sets up application default credentials. I removed some fixtures from `conftest.py` that injected manual credentials from a service account key file.

tseaver (approved):

PR #413, opened 2021-10-26
feat: `to_gbq` uses Parquet by default, use `api_method="load_csv"` for old behavior

### LINKED ISSUE #366: Empty strings inconsistently converted to NULL's when using df.to_gbq()
Here is a reproducible example of inconsistent conversion of empty strings to NULL when exporting to Bigquery.
```
import pandas as pd
from google.cloud import bigquery

df1 = pd.DataFrame.from_dict([{'a': ""}])
df2 = pd.DataFrame.from_dict([{'a': "", 'b': ""}])

df1.to_gbq('aa_temp.df1') # Keeps empty string, as expected.
df2.to_gbq('aa_temp.df2') # Converts empty strings to null values.

client = bigquery.Client()
print(client.list_rows('aa_temp.df1').to_dataframe())
print(client.list_rows('aa_temp.df2').to_dataframe())
```
Actual Output:
```
  a
0  
      a     b
0  None  None
```
Expected Output:
```
  a
0  
      a     b
0    
```
Using python 3.7, pandas 1.2.4, pandas-gbq 0.15.0.

tswast:

Fixes #366 🦕


pandas_gbq/load.py
```diff
@@ -52,45 +54,110 @@ def split_dataframe(dataframe, chunksize=None):
...
+
+
+def load_csv_from_file(
+    client: bigquery.Client,
+    dataframe: pandas.DataFrame,
+    destination_table_ref: bigquery.TableReference,
+    location: Optional[str],
+    chunksize: Optional[int],
+    schema: Optional[Dict[str, Any]],
+):
```
plamut, line 117: There are a lot of common parts with the `load_csv()` function, we can probably extract the common parts out. The difference seems to be just in the extra check if `schema` is `None`, and the actuall for loading a single chunk.

pandas_gbq/load.py
```diff
@@ -52,45 +54,110 @@ def split_dataframe(dataframe, chunksize=None):
...
-                    location=location,
-                ).result()
-            finally:
-                chunk_buffer.close()
+            return load_csv_from_file(
+                client, dataframe, destination_table_ref, location, chunksize, schema
+            )
+    else:
+        raise ValueError(
+            f"got unexpected api_method: {api_method!r}, expected one of 'load_parquet', 'load_csv'"
```
plamut, line 169: (nit) Should the first word in the error message be capitalized?

pandas_gbq/exceptions.py
```diff
@@ -3,6 +3,14 @@
 # license that can be found in the LICENSE file.
 
 
+class GenericGBQException(ValueError):
+    """
+    Raised when an unrecognized Google API Error occurs.
+    """
+
+    pass
```
plamut, line 11: (nit)
`pass` is not needed here, can be removed (just like in `ConversionError`).

pandas_gbq/exceptions.py
```diff
@@ -3,6 +3,14 @@
 # license that can be found in the LICENSE file.
 
 
+class GenericGBQException(ValueError):
```
plamut, line 6: Sanity check - why is `ValueError` chosen as a base, does it really fit all use cases?

I see that other exceptions such as `AccessDenied` also inherit from `ValueError`, which feels somewhat unusual.

plamut (commented):
I can confirm the fix gets rid of the linked issue.

Overall it looks good, the comments are just some nits and one possible refactoring opportunity.

pandas_gbq/exceptions.py
tswast: It's been a ValueError since at least 2015

https://github.com/pandas-dev/pandas/pull/10857/files#diff-ff6e2a94a30a15b176a71736e407806b578e509232bdf7c04e299133a4fb3430R44

I agree that this is unusual. Not sure that we want to be changing it now. Certainly not in this PR.

pandas_gbq/load.py
tswast: I did a bit of refactoring. Indeed, they have more in common than different.

pandas_gbq/load.py
tswast: Done.

I didn't see anything in PEP-8 or the Google style guide regarding capitalization, but makes sense to me to make error messages have proper capitalization.

pandas_gbq/exceptions.py
tswast: Done

pandas_gbq/exceptions.py
plamut: Seems like some voodoo legacy that we indeed should not touch to not awake any sleeping daemons. At least not yet. :)

plamut (approved):
LGTM.

The Python 3.10 check fails, because the BigQuery client does not yet support Python 3.10 and cannot be installed as a dependency.

pandas_gbq/load.py
```diff
@@ -52,45 +56,126 @@ def split_dataframe(dataframe, chunksize=None):
...
-    job_config.allow_quoted_newlines = True
+    job_config.source_format = "PARQUET"
 
-    # Explicit schema? Use that!
     if schema is not None:
         schema = pandas_gbq.schema.remove_policy_tags(schema)
         job_config.schema = pandas_gbq.schema.to_google_cloud_bigquery(schema)
-    # If not, let BigQuery determine schema unless we are encoding the CSV files ourselves.
-    elif not FEATURES.bigquery_has_from_dataframe_with_csv:
-        schema = pandas_gbq.schema.generate_bq_schema(dataframe)
```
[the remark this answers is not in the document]
tswast: Interesting that our tests wouldn't have caught that. Do you have an example of a dataframe that demonstrates this?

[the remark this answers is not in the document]
tswast: FWIW, the reason we don't have this here is that the `google-cloud-bigquery` library does similar dataframe to BQ schema conversion logic if the schema is not populated on the job config: https://github.com/googleapis/python-bigquery/blob/66b3dd9f9aec3fda9610a3ceec8d8a477f2ab3b9/google/cloud/bigquery/client.py#L2625

Issue #426, opened 2021-11-17
support google-cloud-bigquery 3.x (test with pre-release dependencies)


tswast:
There is currently a beta 1 pre-release for google-cloud-bigquery 3.x, it'd be great if we could expand our support for it already -- or at least run some tests to see what breaks.

It does change some default dtypes, so there is a good chance at least some tests will fail.

tswast:
Failing tests with `google-cloud-bigquery==3.0.0b1`

```
nox > Running session prerelease-3.7
nox > Creating virtual environment (virtualenv) using python3.7 in .nox/prerelease-3-7
nox > python -m pip install --extra-index-url https://pypi.fury.io/arrow-nightlies/ --prefer-binary --pre --upgrade pyarrow
nox > python -m pip install --extra-index-url https://pypi.anaconda.org/scipy-wheels-nightly/simple --prefer-binary --pre --upgrade pandas
nox > python -m pip install --prefer-binary --pre --upgrade google-api-core google-cloud-bigquery google-cloud-bigquery-storage google-cloud-core google-resumable-media grpcio
nox > python -m pip install freezegun google-cloud-datacatalog google-cloud-storage google-cloud-testutils IPython mock psutil pytest pytest-cov
nox > python -m pip install db-dtypes google-auth google-auth-oauthlib google-cloud-bigquery google-cloud-bigquery-storage numpy pandas pyarrow pydata-google-auth tqdm
nox > python -m pip install --no-deps -e .[all]
nox > python -m pip freeze
appnope==0.1.2
attrs==21.2.0
backcall==0.2.0
cachetools==4.2.4
certifi==2021.10.8
charset-normalizer==2.0.8
click==8.0.3
coverage==6.2
db-dtypes==0.3.0
decorator==5.1.0
freezegun==1.1.0
google-api-core==2.2.2
google-auth==2.3.3
google-auth-oauthlib==0.4.6
google-cloud-bigquery==3.0.0b1
google-cloud-bigquery-storage==2.10.1
google-cloud-core==2.2.1
google-cloud-datacatalog==3.6.1
google-cloud-storage==1.43.0
google-cloud-testutils==1.2.0
google-crc32c==1.3.0
google-resumable-media==2.1.0
googleapis-common-protos==1.53.0
grpc-google-iam-v1==0.12.3
grpcio==1.42.0
grpcio-status==1.42.0
idna==3.3
importlib-metadata==4.8.2
iniconfig==1.1.1
[...]

PR #428, opened 2021-11-29
test: add session with prerelease dependencies

tswast:

Towards #426  🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -19,7 +19,6 @@
 from __future__ import absolute_import
 import os
 import pathlib
-import re
 import shutil
 
 import nox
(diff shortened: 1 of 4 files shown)


tswast:
Internal configs added in change 412984922. Once submitted, I can trigger the `Kokoro prerelease` job. Currently expected to fail. Won't block merges.

noxfile.py
```diff
@@ -19,7 +19,6 @@
 from __future__ import absolute_import
 import os
 import pathlib
-import re
```
tswast: Oops. Need to add this back via owlbot.py

noxfile.py
```diff
@@ -203,6 +202,89 @@ def docs(session):
     )
 
 
+@nox.session(python=DEFAULT_PYTHON_VERSION)
+def prerelease(session):
```
tswast, line 206: Oops. Added this twice.

plamut (commented):
Looks good, but I'll wait until the external CI issue is resolved to see if OwlBot does not override anything.

tseaver:
@tswast 
>  Once submitted, I can trigger the Kokoro prerelease job. Currently expected to fail. Won't block merges.

Do we even want to run the prerelease checks during a `presubmit`?  ISTM that they are better caught in nightlies.

tswast:
> Do we even want to run the prerelease checks during a presubmit? ISTM that they are better caught in nightlies.

I agree that nightlies are the most useful (assuming flakeybot actually looks at these jobs. I think it should), though for issues like https://github.com/googleapis/python-bigquery-pandas/issues/426 where we know we want to get ahead of some breaking changes, it's nice to have it as a presubmit.

plamut:
Let's see if Owlbot has anything to say now with updated config...

plamut (approved):

PR #442, opened 2021-12-06
fix: allow extreme DATE values such as `datetime.date(1, 1, 1)` in `load_gbq`

### LINKED ISSUE #365: Out of bounds nanosecond timestamp: 1-01-01 00:00:00
Good afternoon.

I have a table where there is a column named maxDate. 

There are some rows (like 1%) that can have the following date: 0001-01-01T00:00:00.

The problem is that it lives in a GBQ from a external vendor to the company, so icannot modify the data and when i call th query like this:
    df = pandas_gbq.read_gbq(
        sql, credentials=credentials)

I have the error:  File "pandas\_libs\tslibs\np_datetime.pyx", line 117, in pandas._libs.tslibs.np_datetime.check_dts_bounds
pandas._libs.tslibs.np_datetime.OutOfBoundsDatetime: Out of bounds nanosecond timestamp: 1-01-01 00:00:00.

Is there a workaround or config so i can execute the query correctly?

Thank you in advance.

### LINKED ISSUE #441: Cannot load dates outside of pandas nanosecond timestamp range with `load_parquet`
https://github.com/googleapis/python-db-dtypes-pandas/issues/45

Possibly we need a fallback for out-of-range values that converts to `object` dtype of datetime.date instead? Or a pandas series containing pyarrow date scalars?

tswast:

Fixes #441 
Towards #365 
🦕

### DIFF
--- a/pandas_gbq/load.py
+++ b/pandas_gbq/load.py
@@ -94,8 +94,13 @@ def cast_dataframe_for_parquet(
             # .astype() with DateDtype. With .astype(), I get the error:
             #
             # TypeError: Cannot interpret '<db_dtypes.DateDtype ...>' as a data type
-            cast_column = pandas.Series(
-                dataframe[column_name], dtype=db_dtypes.DateDtype()
+            cast_column = dataframe[column_name].astype(
+                dtype=db_dtypes.DateDtype(),
+                # Return the original column if there was an error converting
+                # to the dtype, such as is there is a date outside the
+                # supported range.
+                # https://github.com/googleapis/python-bigquery-pandas/issues/441
+                errors="ignore",
             )
         elif column_type in {"NUMERIC", "DECIMAL", "BIGNUMERIC", "BIGDECIMAL"}:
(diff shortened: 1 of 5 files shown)


tswast:
conda-3.7 issue should be resolved by https://github.com/conda-forge/db-dtypes-feedstock/pull/1

pandas_gbq/load.py
```diff
@@ -94,8 +94,13 @@ def cast_dataframe_for_parquet(
             # .astype() with DateDtype. With .astype(), I get the error:
             #
             # TypeError: Cannot interpret '<db_dtypes.DateDtype ...>' as a data type
-            cast_column = pandas.Series(
```
tseaver, line 97: Shouldn't the comment above be removed now?

tseaver (approved):

pandas_gbq/load.py
tswast: Yes. Good call.

PR #443, opened 2021-12-06
feat:  accepts a table ID, which downloads the table without a query

### LINKED ISSUE #266: ENH: Read BigQuery query table without a query
It can cost a lot to run a `SELECT * FROM my_table` query, and yet from an API perspective, this query is completely unnecessary, as one can list the rows from the table directly for free with `tabledata.list`. I propose

```
df = pandas_gbq.read_gbq("my-project.my_dataset.my_table")
```

which will download a table directly (using `bigquery_client.list_rows().to_dataframe()`). Since all queries must contain some kind of whitespace and table IDs cannot contain any whitespace, we can use that to disambiguate query versus table ID.

tswast:

Fixes googleapis/python-bigquery-pandas#266  🦕


pandas_gbq/gbq.py
```diff
@@ -504,13 +532,10 @@ def _download_results(
...
         try:
-            query_job.result()
-            # Get the table schema, so that we can list rows.
-            destination = self.client.get_table(query_job.destination)
-            rows_iter = self.client.list_rows(destination, max_results=max_results)
-
             schema_fields = [field.to_api_repr() for field in rows_iter.schema]
             conversion_dtypes = _bqschema_to_nullsafe_dtypes(schema_fields)
+            # ENDTODO: This is the only difference between table ID and
+
```
tswast, line 538:
```suggestion
```

tests/unit/test_gbq.py
```diff
@@ -505,3 +520,54 @@ def test_read_gbq_calls_tqdm(mock_bigquery_client, mock_service_account_credenti
...
+    df = gbq.read_gbq(
+        "my-project.my_dataset.read_gbq_table",
+        credentials=mock_service_account_credentials,
+        project_id="param-project",
+    )
+    assert df is not None
+
+    mock_bigquery_client.query.assert_not_called()
+    mock_bigquery_client.list_rows.assert_called_with(
+        bigquery.Table("my-project.my_dataset.read_gbq_table"), max_results=None,
```
tswast, line 538: Table comparison by reference-only was not added until a later version of google-cloud-bigquery. Need to update this and the other tests for the `unit-3.7` session

noxfile.py
```diff
@@ -259,7 +259,7 @@ def cover(session):
     test runs (not system test runs), and then erases coverage data.
     """
     session.install("coverage", "pytest-cov")
-    session.run("coverage", "report", "--show-missing", "--fail-under=88")
+    session.run("coverage", "report", "--show-missing", "--fail-under=89")
```
plamut, line 262: I hope this gets bumped to 100% soon, as anything less is susceptible to false positives. For example, refactoring (shortening) a piece of tested code would cause a drop in the overall coverage.

pandas_gbq/gbq.py
```diff
@@ -374,6 +379,26 @@ def process_http_error(ex):
 
         raise GenericGBQException("Reason: {0}".format(ex))
 
+    def download_table(
+        self, table_id, max_results=None, progress_bar_type=None, dtypes=None
+    ):
```
plamut, line 384: (nit) Maybe completely annotate new methods?

tests/unit/test_gbq.py
```diff
@@ -82,6 +83,19 @@ def test__bqschema_to_nullsafe_dtypes(type_, expected):
         assert result == {"x": expected}
 
 
+@pytest.mark.parametrize(
+    ["query_or_table", "expected"],
+    [
+        ("SELECT 1", True),
+        ("dataset.table", False),
+        ("project-id.dataset.table", False),
```
plamut, line 91: I would add some redundant whitespace around table ID's to make sure that doesn't confuse the function, i.e. to test that `strip()` is indeed applied.

plamut (commented):
Looks good, with just a few nits.

noxfile.py
tswast: https://github.com/googleapis/python-bigquery-pandas/issues/392

I plan to tackle this next, as I do hope the next pandas-gbq release can be a 1.0 release, but wouldn't be comfortable without full test coverage.

pandas_gbq/gbq.py
tswast: Good call, done.

I probably could have used `Dict[str, Any]` like we did [here](https://github.com/googleapis/python-bigquery/blob/66b3dd9f9aec3fda9610a3ceec8d8a477f2ab3b9/google/cloud/bigquery/job/query.py#L1561) for dtypes, but maybe something more specific would be better?

Note: we aren't currently running any sort of type checking in this repo. There's an open issue for it here: https://github.com/googleapis/python-bigquery-pandas/issues/325

plamut: For a start, `Dict[str, any]` is probably just fine, especially if type checks are not implemented yet. Let's just make sure it's consistent with the docstring, the latter currently says `Optional[Map[str, Union[str, pandas.Series.dtype]]]`.

pandas_gbq/gbq.py
```diff
@@ -7,9 +7,17 @@
...
+import typing
+from typing import Dict, Optional, Union
 
 import numpy as np
 
+# Only import at module-level at type checking time to avoid circular
+# dependencies in the pandas package, which has an optional dependency on
+# pandas-gbq.
+if typing.TYPE_CHECKING:
+    import pandas
```
plamut, line 19: This will never be executed during the tests (`TYPE_CHECKING` is `False` at runtime), thus at some point coverage will complain. Let's add a `# pragma: NO COVER` comment to ignore this block.

plamut (approved):
LGTM.

The two comments are non-blocking.
(making the annotation consistent with the docstring and excluding the TYPE_CHECKING imports from coverage)

tswast:
Thanks for the review @plamut!

PR #444, opened 2021-12-06
fix: `read_gbq` supports extreme DATETIME values such as `0001-01-01 00:00:00`

### LINKED ISSUE #365: Out of bounds nanosecond timestamp: 1-01-01 00:00:00

tswast:

Fixes #365  🦕


tswast:
I think we need to remove the `datetime64[ns]` values from the `dtypes` map, since they don't map to the full allowed range of values in BigQuery. Instead, do a post-processing step with a couple of fallbacks depending on out-of-bounds errors (assuming the user hasn't already manually overridden the dtypes).

https://github.com/googleapis/python-bigquery-pandas/blob/e13abaf015cd1ea9da3ad5063680bf89e18f0fac/pandas_gbq/gbq.py#L583

ci/requirements-3.7-0.24.2.conda
```diff
@@ -1,10 +1,11 @@
 codecov
 coverage
-db-dtypes==0.3.0
+db-dtypes==0.3.1
 fastavro
 flake8
 numpy==1.16.6
-google-cloud-bigquery==1.11.1
+google-cloud-bigquery==1.26.1
```
tswast, line 7: Needed for `date_as_object` parameter

pandas_gbq/gbq.py
```diff
@@ -617,23 +619,18 @@ def _bqschema_to_nullsafe_dtypes(schema_fields):
...
     """
+    import db_dtypes
+
     # If you update this mapping, also update the table at
     # `docs/reading.rst`.
     dtype_map = {
-        "DATE": "datetime64[ns]",
-        "DATETIME": "datetime64[ns]",
         "FLOAT": np.dtype(float),
-        "GEOMETRY": "object",
```
tseaver, line 626: I'm not sure how the changes here to non-datetime-related mappings relates to this PR.

If these changes are intentional, then the comment above seems to require a corresponding update to `docs/reading.rst`.

setup.py
```diff
@@ -28,12 +28,13 @@
     "pandas >=0.24.2",
     "pyarrow >=3.0.0, <7.0dev",
     "pydata-google-auth",
-    "google-api-core >=1.14.0",
+    "google-api-core >=1.21.0",
```
tseaver, line 31: Seems unrelated, and unmotivated by anything in the [changelog for that release](https://github.com/googleapis/python-api-core/releases/tag/v1.21.0).

testing/constraints-3.7.txt
```diff
@@ -6,10 +6,10 @@
 # e.g., if setup.py has "foo >= 1.14.0, < 2.0.0dev",
 # Then this file should have foo==1.14.0
 db-dtypes==0.3.1
-google-api-core==1.14.0
-google-auth==1.4.1
+google-api-core==1.21.0
+google-auth==1.18.0
```
tseaver, line 10: Doesn't match the minimum constraint in `setup.py`.

setup.py
tswast: Correct. Needed to update due to updating the minimum `google-cloud-bigquery`, though.

We do use `google-api-core` directly, so I think it makes sense to include here still.

pandas_gbq/gbq.py
tswast: `object` types were removed because it's the default anyway.

testing/constraints-3.7.txt
tswast: Updated `setup.py`. Needed this version due to updated `google-api-core` (via `google-cloud-bigquery`)

shollyman (approved):
LGTM, though all this version checking feels mildly terrifying.

pandas_gbq/gbq.py
```diff
@@ -656,28 +653,43 @@ def _bqschema_to_nullsafe_dtypes(schema_fields):
...
+    # `docs/reading.rst`.
+    dtype_map = {
+        "DATE": db_dtypes.DateDtype(),
+        "DATETIME": "datetime64[ns]",
+        "TIMESTAMP": "datetime64[ns]",
+    }
 
     for field in schema_fields:
-        column = str(field["name"])
+        # This method doesn't modify ARRAY/REPEATED columns.
```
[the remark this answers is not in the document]
tswast: Potential TODO, but such a low priority I don't think it's worth calling out. Now that we have https://github.com/googleapis/python-db-dtypes-pandas we have more flexibility in terms of creating dtypes that are more efficient than Python object columns. Though in this case, I'm not sure we'd have any better of an approach than https://github.com/xhochy/fletcher

tswast:
> LGTM, though all this version checking feels mildly terrifying.

Yeah, for sure... I'd very much like to give our pandas-gbq users as wide a set of versions as possible, though. Those folks are often stuck in (notebook) environments with some core dependencies locked.

PR #451, opened 2021-12-24
chore: update release_level in repo-metadata.json

parthea:
Update .repo-metadata.json as required by go/library-data-integrity


.repo-metadata.json
```diff
@@ -2,14 +2,15 @@
     "name": "pandas-gbq",
     "name_pretty": "Google BigQuery connector for pandas",
     "product_documentation": "https://cloud.google.com/bigquery",
-    "client_documentation": "https://pandas-gbq.readthedocs.io/en/latest/",
+    "client_documentation": "https://cloud.google.com/python/docs/reference/pandas-gbq/latest",
```
tswast, line 5: `pandas-gbq` isn't under Cloud RAD because it's not a client library (also needs intersphinx to work well)

```suggestion
    "client_documentation": "https://googleapis.dev/python/pandas-gbq/latest/",
```

.repo-metadata.json
```diff
@@ -2,14 +2,15 @@
...
+    "release_level": "preview",
     "language": "python",
     "library_type": "INTEGRATION",
     "repo": "googleapis/python-bigquery-pandas",
     "distribution_name": "pandas-gbq",
     "api_id": "bigquery.googleapis.com",
     "default_version": "",
-    "codeowner_team": "@googleapis/api-bigquery"
+    "codeowner_team": "@googleapis/api-bigquery",
+    "api_shortname": "pandas-gbq"
```
tswast, line 15: The docs you like to say this field is "required if library_type is GAPIC*. Optional for others."

Also, wouldn't `api_shortname` be just `bigquery` so that labeling and such maps to the correct backend API?

tswast (changes requested):

parthea:
You're right. Since it's optional, I've removed it to avoid confusion.

.repo-metadata.json
parthea: You're right. Since it's optional, I've removed it to avoid confusion.

tswast (approved):

PR #539, opened 2022-06-15
fix: allow `to_gbq` to run without `bigquery.tables.create` permission.

### LINKED ISSUE #538: `to_gbq` requires `bigquery.tables.create` permission even with if_exists="append"
I have a workload that runs with a service account that has permissions only to write to an existing table but not to create one or change the schema.  I encountered a problem that even though the table exists and we passed the argument if_exists="append" we got a `403` error that the service account doesn't have `bigquery.tables.create` permission, though it shouldn't be needed at all.

After investigating it I found the problem. In the LoadJobConfig sent with the request [there is a field](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationLoad) `createDisposition` that if it's not set the default is  `CREATE_IF_NEEDED`  and that  requires the service account which submits the jobs to have  the `bigquery.tables.create` permission.

I suggest setting the field `createDisposition` to `CREATE_NEVER` to prevent the need to have this permission.

There is no fear that this will restrict users who want `to_gbq` to create the table for them if it doesn't exist because at the state the LoadJobConfig is created the table **must** already exists.  

#### Steps to reproduce

  1. Create a service account with these permissions
     - roles/bigquery.dataViewer
     -  custom role with:
       bigquery.datasets.get
       bigquery.tables.get
       bigquery.tables.updateData

  
  2. Create a table
  3.  Run this code 
```python
dataframe = pandas.DataFrame({"test": ["some test data"]})

pandas_gbq.to_gbq(
    dataframe=dataframe,
    destination_table="my_dataset.my_table_name",
    project_id="my_project_id",
    if_exists="append",
)
```

#### Stack trace
```
    pandas_gbq.to_gbq(
  File "/lib/python3.10/site-packages/pandas_gbq/gbq.py", line 1198, in to_gbq
    connector.load_data(
  File "/lib/python3.10/site-packages/pandas_gbq/gbq.py", line 610, in load_data
    self.process_http_error(ex)
  File "/lib/python3.10/site-packages/pandas_gbq/gbq.py", line 386, in process_http_error
pandas_gbq.exceptions.GenericGBQException: Reason: 403 POST https://bigquery.googleapis.com/upload/bigquery/v2/projects/my_project_id/jobs?uploadType=multipart: Access Denied: Dataset my_project_id:my_dataset: Permission bigquery.tables.create denied on dataset my_project_id:my_dataset (or it may not exist).
```

acarmel:
Fixes googleapis/python-bigquery-pandas#538  🦕

### DIFF
--- a/pandas_gbq/load.py
+++ b/pandas_gbq/load.py
@@ -119,6 +119,7 @@ def load_parquet(
 ):
     job_config = bigquery.LoadJobConfig()
     job_config.write_disposition = "WRITE_APPEND"
+    job_config.create_disposition = "CREATE_NEVER"
     job_config.source_format = "PARQUET"
 
     if schema is not None:
@@ -148,6 +149,7 @@ def load_csv(
 ):
     job_config = bigquery.LoadJobConfig()
     job_config.write_disposition = "WRITE_APPEND"
+    job_config.create_disposition = "CREATE_NEVER"
     job_config.source_format = "CSV"
     job_config.allow_quoted_newlines = True
 


tswast (approved):
Thanks for the investigation and the fix!

Note: this might make it harder to create tables via the load job as I had planned in https://github.com/googleapis/python-bigquery-pandas/issues/425 but we can address that if/when we try to do that.

acarmel:
> Thanks for the investigation and the fix!
> 
> Note: this might make it harder to create tables via the load job as I had planned in #425 but we can address that if/when we try to do that.

Thanks!

In the current design you already create the table  before the loading.
https://github.com/googleapis/python-bigquery-pandas/blob/439d5888550f1ad11418b65fe84f3f48434d3515/pandas_gbq/gbq.py#L1153-L1191)


And when you load  you create your own JobConfig. 
https://github.com/googleapis/python-bigquery-pandas/blob/439d5888550f1ad11418b65fe84f3f48434d3515/pandas_gbq/load.py#L120-L122

I guess that once you add the option to let the user provide config then you can go with that config and not create one of yours and otherwise create a default config with reasonable default.

steffnay (approved):

PR #564, opened 2022-09-01
test: adds exclusion: prerelease grpcio!=1.49.0rc0

chalmerlowe:
Provides an exclusion to avoid using a problematic version of grpcio.


dandhlee:
Owlbot's reverting the changes, see https://github.com/googleapis/python-bigquery-pandas/blob/388196a9fe7bcf954bdae258ad7b79081b1dfc86/owlbot.py#L77-L79 for example on how to make changes you'd like.

PR #581, opened 2022-10-25
feat: add WriteDisposition to replace "if_exists='fail'/'replace'/'append'" behavior

### LINKED ISSUE #118: CLN: Use WriteDisposition for if_exists=fail/replace/append behavior in to_gbq
Now that to_gbq is using load jobs instead of streaming to write to table, it could use the write disposition parameter to perform this behavior.

I believe by doing so we would unlock https://github.com/pydata/pandas-gbq/issues/43 since using write disposition could replace just one partition, but the current logic of deleting and recreating a table cannot handle partitions.

aribray:
Use `LoadJobConfig.WriteDisposition` to replace `if_exists='fail'/'replace'/'append'` behavior in `to_gbq()`

The WriteDisposition value specifies the action that occurs if destination table already exists.

- if_exists="fail" ----> write_disposition="WRITE_EMPTY" 
- if_exists="append" ----> write_disposition="WRITE_APPEND"
- if_exists="replace" ----> write_disposition="WRITE_TRUNCATE"

By using WriteDisposition, we can move away from the `delete_and_recreate_table()` behavior

Fixes #118 #43 #453


pandas_gbq/gbq.py
```diff
@@ -1001,15 +998,7 @@ def to_gbq(
     reauth : bool, default False
         Force Google BigQuery to re-authenticate the user. This is useful
         if multiple accounts are used.
-    if_exists : str, default 'fail'
```
tswast, line 1004: This is a breaking change, right?

I figured we could pretty easily map from `if_exists` to the corresponding write disposition without breaking our users.

PR #583, opened 2022-10-27
feat: map "if_exists" value to LoadJobConfig.WriteDisposition

### LINKED ISSUE #118: CLN: Use WriteDisposition for if_exists=fail/replace/append behavior in to_gbq

aribray:
Use `LoadJobConfig.WriteDisposition` to replace `if_exists='fail'/'replace'/'append'` behavior in `to_gbq()`

The WriteDisposition value specifies the action that occurs if destination table already exists.

- if_exists="fail" ----> write_disposition="WRITE_EMPTY" 
- if_exists="append" ----> write_disposition="WRITE_APPEND"
- if_exists="replace" ----> write_disposition="WRITE_TRUNCATE"

By using WriteDisposition, we can move away from the `delete_and_recreate_table()` behavior

Fixes #118 #43 #453


pandas_gbq/load.py
```diff
@@ -113,20 +113,19 @@ def load_parquet(
...
     dataframe: pandas.DataFrame,
     destination_table_ref: bigquery.TableReference,
+    write_disposition: str,
     location: Optional[str],
     schema: Optional[Dict[str, Any]],
     billing_project: Optional[str] = None,
 ):
     job_config = bigquery.LoadJobConfig()
-    job_config.write_disposition = "WRITE_APPEND"
-    job_config.create_disposition = "CREATE_NEVER"
```
aribray, line 122: The `create_disposition` attribute defaults to `CREATE_IF_NEEDED`

ci/requirements-3.7-0.24.2.conda
```diff
@@ -1,14 +1,17 @@
 codecov
```
aribray, line 1: Pinning dependencies was causing a bunch of problems for the CircleCI (3.7) build. We should move these tests over to Kokoro

aribray:
@tswast are the version bumps/PR description added to the Changelog automatically?

tswast:
> @tswast are the version bumps/PR description added to the Changelog automatically?

No, that needs to be included in the commit message when we 'squash and merge'. See: https://github.com/googleapis/release-please#what-if-my-pr-contains-multiple-fixes-or-features

pandas_gbq/load.py
```diff
@@ -234,13 +236,15 @@ def load_chunks(
     schema=None,
     location=None,
     api_method="load_parquet",
+    write_disposition="WRITE_EMPTY",
```
rosiezou, line 239: nit: define a global string constant for this arg, same with `api_method`

pandas_gbq/gbq.py
```diff
@@ -1114,15 +1109,21 @@ def to_gbq(
...
-    if if_exists not in ("fail", "replace", "append"):
-        raise ValueError("'{0}' is not valid for if_exists".format(if_exists))
-
     if "." not in destination_table:
         raise NotFoundException(
             "Invalid Table Name. Should be of the form 'datasetId.tableId' or "
             "'projectId.datasetId.tableId'"
         )
 
+    if if_exists not in ("fail", "replace", "append"):
```
shollyman, line 1118: minor consideration:  Should we case coerce `if_exists` to lowercase when evaluating, or throw the value error unless they're exact match?  I'm fine with the strictness approach, but something to consider.

shollyman (approved):
Thanks for putting this together.  Switching to leveraging write disposition cleans things up nicely.

pandas_gbq/gbq.py
aribray: That's a good point. I'll make a note to do that

PR #595, opened 2022-12-08
feat: adds ability to provide redirect uri

chalmerlowe:
WIP PR for discussion: aiming to provide the ability to include a redirect URI, client ID, and client secrets to facilitate the migration away from "out of band" OAuth authentication.

@tswast 

For context, there are changes in these repos:
* https://github.com/googleapis/python-bigquery-pandas/pull/595 #python-bigquery-pandas
* https://github.com/pydata/pydata-google-auth/pull/58 (Non-google repository)

### DIFF
--- a/docs/_static/js/authcodescripts.js
+++ b/docs/_static/js/authcodescripts.js
@@ -1,3 +1,10 @@
+
+
+// Copyright (c) 2023 pandas-gbq Authors All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+
 function onloadoauthcode() {
     const PARAMS = new Proxy(new URLSearchParams(window.location.search), {
         get: (searchParams, prop) => searchParams.get(prop),
--- a/docs/_static/js/authcodescripts.js
+++ b/docs/_static/js/authcodescripts.js
@@ -1,0 +1,31 @@
+function onload() {
+    alert("Hello! DINOSAUR!");
+    const PARAMS = new Proxy(new URLSearchParams(window.location.search), {
+        get: (searchParams, prop) => searchParams.get(prop),
+    });
+    const AUTH_CODE = PARAMS.code;
+
+    document.querySelector('.auth-code').textContent = AUTH_CODE;
+
+    setupCopyButton(document.querySelector('.copy'), AUTH_CODE);
+}
+
+function setupCopyButton(button, text) {
+    button.addEventListener('click', () => {
+        navigator.clipboard.writeText(text);
+        button.textContent = "Verification Code Copied";
+        setTimeout(() => {
+            // Remove the aria-live label so that when the
+            // button text changes back to "Copy", it is
+            // not read out.
+            button.removeAttribute("aria-live");
+            button.textContent = "Copy";
+        }, 1000);
+
+        // Re-Add the aria-live attribute to enable speech for
+        // when button text changes next time.
+        setTimeout(() => {
+            button.setAttribute("aria-live", "assertive");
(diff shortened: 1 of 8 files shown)


chalmerlowe:
These failing tests are due to the fact that pydata-google-auth also needs a matching update and that PR (noted above) has not been merged into the main project yet. When that is merged, the expectation is that these tests will pass.

docs/_static/js/authcodescripts.js
```diff
@@ -0,0 +1,30 @@
+function onloadoauthcode() {
```
tswast, line 1: Nit: We'll need a license header here. pandas-gbq inherited a BSD license, so it'll look a bit different from the Apache 2.0 header.

```suggestion


// Copyright (c) 2023 pandas-gbq Authors All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.


function onloadoauthcode() {
```

pandas_gbq/auth.py
```diff
@@ -43,10 +49,11 @@ def get_credentials(
 
     credentials, default_project_id = pydata_google_auth.default(
         SCOPES,
-        client_id=CLIENT_ID,
-        client_secret=CLIENT_SECRET,
+        client_id=client_id,
```
tswast, line 52: We need to keep our default client ID and secret.

Note: if `auth_local_webserver = False`, then ideally we provide a different default client ID & secret & a default redirect URI, but we can/should do that in a follow-up PR.

pandas_gbq/gbq.py
```diff
@@ -864,6 +877,8 @@ def read_gbq(
         or
         :func:`google.oauth2.service_account.Credentials.from_service_account_file`
         instead.
+    # TODO: add parameters for auth_redirect_uri, client_id, client_secret
```
tswast, line 880: Need to complete this TODO

pandas_gbq/gbq.py
```diff
@@ -912,6 +927,9 @@ def read_gbq(
         credentials=credentials,
         private_key=private_key,
         use_bqstorage_api=use_bqstorage_api,
+        auth_redirect_uri=auth_redirect_uri,
```
tswast, line 930: There are other uses of `GbqConnector` that need to be updated. Please add these auth argument to `to_gbq` as well.

tswast (changes requested):

pandas_gbq/gbq.py
chalmerlowe: done

pandas_gbq/gbq.py
chalmerlowe: Done. Also added the same docstrings to the `to_gbq()` function.

pandas_gbq/gbq.py
```diff
@@ -864,6 +877,15 @@ def read_gbq(
         or
         :func:`google.oauth2.service_account.Credentials.from_service_account_file`
         instead.
+    auth_redirect_uri : str
+        Path to the authentication page for organization-specific authentication
+        workflows.
```
tswast, line 882:
```suggestion
        workflows. Used when `auth_local_webserver=False`.
```

pandas_gbq/gbq.py
```diff
@@ -1071,6 +1099,15 @@ def to_gbq(
...
```
tswast, line 1104:
```suggestion
        workflows. Used when `auth_local_webserver=False`.
```

tswast (approved):

tswast (approved):

PR #597, opened 2022-12-21
chore: updates python version to 3.11

chalmerlowe:
Introducing this for discussion. Planning on adding version 3.11 as a test platform. Not sure about all the inner workings behind our codebase, so there are prolly a number of things that need to be added/edited. Feel free to comment.

@tswast @parthea

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


chalmerlowe:
@parthea 
@swast

what else do we need to do to manage upgrading/updating the allowed Python versions?

tswast:
Owlbot needs to run to make sure the templated files (such as the noxfile) are updated. I've added the label, so hopefully it should swing by and add a commit.

parthea (approved):

PR #601, opened 2023-01-13
bug: adds folder ref for javascript files

chalmerlowe:
Updates the javascript folder reference by adding a line to the `conf.py` file.


tswast (approved):

chalmerlowe:
My recollection is that this is no longer necessary and was superceded by #595

meredithslota:
There are in fact 0 changes in this PR, so closing based on that and @chalmerlowe's comment above.

PR #603, opened 2023-01-20
feat: updates the user instructions re OAuth

chalmerlowe:
Provides user instructions as part of the pandas-gbg documentation to support authentication using OAuth for highly constrained development environments.

BEGIN_COMMIT_OVERRIDE
docs: updates the user instructions re OAuth
END_COMMIT_OVERRIDE


tswast (commented):
Content LGTM, but I think this might be better on a separate page so that we don't overwhelm folks for whom the default flow works fine.

chalmerlowe:
@tswast 

> Content LGTM, but I think this might be better on a separate page so that we don't overwhelm folks for whom the default flow works fine.

Good point. I separated out the content into two pages. I added cross references within each document to help ensure that users are guided to the appropriate content.

tswast (approved):

PR #629, opened 2023-05-09
fix: add exception context to GenericGBQExceptions

quoimec:
The `from` keyword adds a `__cause__` parameter to the thrown exception. This allows external applications access to the originally thrown exception.

My required usecase: my application has experienced BQ rate limiting errors but I am only getting GenericGBQExceptions out the other side, this change will allow me to be more nuanced with my error handling and apply rate limiting pauses only when appropriate.


tswast:
Thanks, @quoimec ! This LGTM. Any thoughts on how we can unit test this?

tswast (approved):

PR #699, opened 2023-11-15
chore: include api error info in raised errors

Linchin:
Part of fix for internal issue 219787554. This will make our error messages more actionable for users.

Fixes #<issue_number_goes_here> 🦕

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -486,13 +486,12 @@ class GbqConnector(object):
         except (RefreshError, ValueError) as ex:
             if self.private_key:
                 raise AccessDenied(
-                    "The service account credentials are not " f"valid: {ex}"
+                    f"The service account credentials are not valid: {ex}"
                 )
             else:
                 raise AccessDenied(
                     "The credentials have been revoked or expired, "
-                    "please re-run the application to re-authorize: "
-                    f"{ex}"
+                    f"please re-run the application to re-authorize: {ex}"
                 )
         except self.http_error as ex:
             self.process_http_error(ex)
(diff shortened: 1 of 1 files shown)


pandas_gbq/gbq.py
```diff
@@ -483,13 +483,16 @@ def run_query(self, query, max_results=None, progress_bar_type=None, **kwargs):
                 project=self.project_id,
             )
             logger.debug("Query running...")
-        except (RefreshError, ValueError):
+        except (RefreshError, ValueError) as ex:
             if self.private_key:
-                raise AccessDenied("The service account credentials are not valid")
+                raise AccessDenied(
+                    "The service account credentials are not " f"valid: {ex}"
```
shollyman, line 489: This string is short enough, is it worth avoid the concat and just using the whole thing as a format string?

pandas_gbq/gbq.py
```diff
@@ -483,13 +483,16 @@ def run_query(self, query, max_results=None, progress_bar_type=None, **kwargs):
...
+                )
             else:
                 raise AccessDenied(
                     "The credentials have been revoked or expired, "
-                    "please re-run the application to re-authorize"
+                    "please re-run the application to re-authorize: "
```
shollyman, line 494: similar feedback here on format string here as well

shollyman (approved):
LGTM, with some minor feedback on the string construction

pandas_gbq/gbq.py
Linchin: Good catch! I forgot to double check after the black reformat.

pandas_gbq/gbq.py
Linchin: corrected

PR #701, opened 2023-11-21
feat: add 'columns' as an alias for 'col_order'

### LINKED ISSUE #617: feat: add "columns" as an alias for "col_order"
One thing I'd like is to align `read_gbq` with `read_sql` and `read_sql_table`. Those methods call what we call "col_order", "columns". See: https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html

kiraksi:
This PR
- adds 'columns' as an alias for 'col_order'
- updates documentation to prefer 'columns' so that we can choose columns we want to return, and not return entire order
- adds tests to test this functionality and to gracefully fail if a user supplies both

- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery-pandas/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes #617 🦕


docs/reading.rst
```diff
@@ -28,7 +28,8 @@ destination DataFrame as well as a preferred column order as follows:
        'SELECT * FROM `test_dataset.test_table`',
        project_id=projectid,
        index_col='index_column_name',
-       col_order=['col1', 'col2', 'col3'])
+       col_order=['col1', 'col2', 'col3'],
+       columns=['col1', 'col2'])
```
tswast, line 32: This doesn't belong in the code sample. The user should only use one (prefer `columns`)

pandas_gbq/gbq.py
```diff
@@ -964,6 +967,9 @@ def read_gbq(
                 'Index column "{0}" does not exist in DataFrame.'.format(index_col)
             )
 
+    # Creating an alias for col_order, which is columns
```
chalmerlowe, line 970: Do we need a check for whether the user provides both `col_order` AND `columns`, with a warning OR error message to say that only one is allowed?

tests/system/test_gbq.py
```diff
@@ -600,6 +600,49 @@ def test_tokyo(self, tokyo_dataset, tokyo_table, project_id):
...
+
+        correct_frame = DataFrame(
+            {"string_1": ["a"], "string_2": ["b"], "string_3": ["c"]}
+        )[columns]
+
+        # Verify that the result_frame matches the expected DataFrame
+        tm.assert_frame_equal(result_frame, correct_frame)
+
+        # Ensure that the order of columns in the result_frame matches the specified order
+        assert sorted(columns) == sorted(result_frame.columns)
```
chalmerlowe, line 644: I think I am missing something. 
Do we want both the result and the expected values to be `sorted()`?

pandas_gbq/gbq.py
kiraksi: I talked to Tim and he mentioned that we should probably fail with ValueError if both are provided, will implement this in next commit

tests/system/test_gbq.py
kiraksi: If I am comparing the results, wouldn't I need to sort them so that the order doesn't effect the comparison?

chalmerlowe: The comment says: 

`# Ensure that the order of columns in the result_frame matches the specified order`

Presuming that the "specified order" is the order that the user lists the columns when they include the `columns` argument in the function call:

`columns = ["col1", "col9", "col2"]`

If we sort `columns` then we are destroying the specified order so we can't check to see if the columns in result_frame matches that order.

`columns = sorted(["col1", "col9", "col2"])`
`columns = ["col1", "col2", "col9"]`

Similarly, if we sort both `columns` and `result_frame` then we are reorganizing both of them and then we lose all info about the specified order in each of them and end up only testing whether the two sequences have the exact same elements but can't tell if we matched the originally specified order.

`columns = sorted(["col1", "col9", "col2"])       # specified order`
`result_frame = sorted(["col1", "col9", "col2"])`

`columns = ["col1", "col2", "col9"].              # no longer in specified order`
`result_frame = sorted(["col1", "col2", "col9"])`

It seems to me, that if we want to match the specified order of the columns in the `columns` variable, then we shouldn't be sorting either of these.

`columns = ["col1", "col9", "col2"]`
`result_frame = ["col1", "col9", "col2"]`

`columns == result_frame`
`True`

pandas_gbq/gbq.py
```diff
@@ -733,6 +733,7 @@ def read_gbq(
     project_id=None,
     index_col=None,
     col_order=None,
+    columns=None,
```
tswast, line 736: I don't think it matters too much here, but technically this is a breaking change. Python allows these arguments to be passed in positionally or by keyword unless we use some special syntax to force them to use it as a keyword argument.

pandas_gbq/gbq.py
```diff
@@ -774,6 +775,8 @@ def read_gbq(
     col_order : list(str), optional
         List of BigQuery column names in the desired order for results
         DataFrame.
+    columns : list(str), optional
+        Alias for col_order
```
tswast, line 779: Let's flip this. Add the description for col_order here and the description of col_order should be: Alias for columns, retained for backwards compatibility.

pandas_gbq/gbq.py
```diff
@@ -964,10 +967,16 @@ def read_gbq(
...
+        columns = col_order
+    elif col_order and columns:
+        raise ValueError("Must specify either columns or col_order, not both")
+
     # Change the order of columns in the DataFrame based on provided list
-    if col_order is not None:
-        if sorted(col_order) == sorted(final_df.columns):
-            final_df = final_df[col_order]
+    if columns is not None:
+        if sorted(columns) == sorted(final_df.columns):
```
tswast, line 978: This is a difference between how we're using `columns` in BigQuery DataFrames and pandas-gbq. We expect columns to be a subset of all columns in the table.

Please add a TODO to allow this to be a subset instead.

In a follow-up PR, please pass columns to download_table so that we can only download a subset of columns.

https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_list_rows has a selected_fields argument for this purpose.

tswast (changes requested):

chalmerlowe (approved):
Beyond the changes Tim mentioned, LGTM.
Approved pending needed resolution of Tim's comments.

pandas_gbq/gbq.py
```diff
@@ -966,10 +970,17 @@ def read_gbq(
                 'Index column "{0}" does not exist in DataFrame.'.format(index_col)
             )
 
+    # Using columns as an alias for col_order, raising an error if both provided
+    if col_order and not columns:
+        columns = col_order
+    elif col_order and columns:
+        raise ValueError("Must specify either columns or col_order, not both")
```
chalmerlowe, line 977:
```suggestion
        raise ValueError("Must specify either columns (preferred) or col_order, not both")
```

tswast (approved):
Thanks!

Issue #710, opened 2023-12-08
use faster `query_and_wait` method when available to skip BQ Storage API if all query results are already available


tswast:
Similar to https://github.com/googleapis/python-bigquery/issues/1745, pandas-gbq should benefit from the optimizations done in https://github.com/googleapis/python-bigquery/issues/589.

This might require a progress bar feature in `query_and_wait` if we want to keep the progress bar for waiting for the query to finish, but we need to be careful when introducing such a progress bar to avoid timing out API requests too soon.

PR #720, opened 2024-01-09
refactor: move query and wait logic to separate module

tswast:
This prepares the way for using the `query_and_wait` method built-in to the client library when available.

Towards #710 
🦕


tswast:
Conda test failure appears unrelated and tracked here: https://github.com/googleapis/python-bigquery-pandas/issues/708

pandas_gbq/exceptions.py
```diff
@@ -35,3 +35,12 @@ class PerformanceWarning(RuntimeWarning):
...
     """
+
+
+class QueryTimeout(ValueError):
+    """
+    Raised when the query request exceeds the timeoutMs value specified in the
+    BigQuery configuration.
+    """
+
+    pass
```
Linchin, line 46: nit: should we keep the `pass` here? Other error classes do not have this.

pandas_gbq/query.py
```diff
@@ -0,0 +1,135 @@
...
+
+def _wait_for_query_job(
+    connector,
+    client: bigquery.Client,
+    query_reply: bigquery.QueryJob,
+    timeout_ms: Optional[float],
+):
+    """Wait for query to complete, pausing occasionally to update progress.
+
+    Args:
```
Linchin, line 42: I see this method has been moved out to be a helper function, maybe we should also update the description as the parameters have changed.

Linchin (approved):
Thank you @tswast! All looks good except for some small comments.

pandas_gbq/exceptions.py
tswast: You're correct. `pass` is _not_ needed because there is a docstring. I'll remove.

pandas_gbq/query.py
tswast: Updated docstrings here and added docstrings to the `query_and_wait()` method.

PR #721, opened 2024-01-09
fix: update runtime check for min google-cloud-bigquery to 3.3.5

tswast:
The minimum version of google-cloud-bigquery was updated to 3.3.5 in pandas-gbq
version 0.18.0 (released November 2022). This change updates the runtime check
in features.py to reflect that minimum version and removes some dead code for
feature checks that are no longer relevant.

~Please review #720 first.~ Edit: removed conflicting setup.py change so these changes are now independent.

🦕


Linchin (approved):
Thank you! All looks good.

PR #722, opened 2024-01-09
feat: use faster query_and_wait method from google-cloud-bigquery when available

### LINKED ISSUE #710: use faster `query_and_wait` method when available to skip BQ Storage API if all query results are already available

tswast:

Requires #720, #721, and https://github.com/googleapis/python-bigquery/pull/1793 to be merged first.

Fixes #710 🦕

### DIFF
--- a/tests/system/test_gbq.py
+++ b/tests/system/test_gbq.py
@@ -426,33 +426,6 @@ class TestReadGBQIntegration(object):
         )
         tm.assert_frame_equal(df, DataFrame({"valid_string": ["PI"]}))
 
-    def test_configuration_without_query(self, project_id):
-        sql_statement = "SELECT 1"
-        config = {
-            "copy": {
-                "sourceTable": {
-                    "projectId": project_id,
-                    "datasetId": "publicdata:samples",
-                    "tableId": "wikipedia",
-                },
-                "destinationTable": {
-                    "projectId": project_id,
-                    "datasetId": "publicdata:samples",
-                    "tableId": "wikipedia_copied",
-                },
-            }
-        }
-        # Test that only 'query' configurations are supported
-        # nor 'copy','load','extract'
-        with pytest.raises(ValueError):
-            gbq.read_gbq(
-                sql_statement,
-                project_id=project_id,
-                credentials=self.credentials,
-                configuration=config,
-                dialect="legacy",
-            )
-
     def test_configuration_raises_value_error_with_multiple_config(self, project_id):
         sql_statement = "SELECT 1"
         config = {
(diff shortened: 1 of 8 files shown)


tswast:
I tested this manually in a local ipython instance to verify that the latency is improved when using this new functionality:

**After** this change:

```python
# In [1]:
import pandas_gbq

# In [2]: %%timeit -n10 -r10
df = pandas_gbq.read_gbq("SELECT 1", progress_bar_type=None)

# Out [2]:
# 394 ms ± 37.8 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
```

**Before** this change:

```python
# In [1]:
import pandas_gbq

# In [2]: %%timeit -n10 -r10
df = pandas_gbq.read_gbq("SELECT 1", progress_bar_type=None)

# Out [2]:
# 1.08 s ± 102 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)
```

This is quite similar to the results I observed when implementing this optimization in the DB-API connector. https://github.com/googleapis/python-bigquery/pull/1747#issuecomment-1852368885

tests/system/test_gbq.py
```diff
@@ -426,33 +426,6 @@ def test_query_inside_configuration(self, project_id):
         )
         tm.assert_frame_equal(df, DataFrame({"valid_string": ["PI"]}))
 
-    def test_configuration_without_query(self, project_id):
```
tswast, line 429: I'm not 100% sure why this isn't failing anymore. I guess jobs.query is more lenient about unknown parameters than jobs.insert is?

Linchin: Could you share a link to a failing test log? I wasn't able to find it in test fusion.

Linchin: Should we keep this test?

Linchin:
All looks good except for that I wanted to double check on the removed system test.

tests/system/test_gbq.py
tswast: I tested locally with https://github.com/googleapis/python-bigquery/pull/1793 and the request body now includes unknown parameters, but surprisingly the REST API seems to ignore these. I'll follow-up on that PR to raise on common errors, such as passing in a configuration object of the wrong type.

tswast: Updated https://github.com/googleapis/python-bigquery/pull/1793 with some client-side checks for common mistakes of sending the wrong type for `job_config` and now this test is passing.

Will need to wait for a `google-cloud-bigquery` release with that fix before merging this PR.

tswast:
@Linchin Tests are passing after the most recent google-cloud-bigquery release. OK for another look.

Linchin:
LGTM, thank you!

Linchin (approved):

PR #723, opened 2024-01-11
chore: pip check on conda forge dependencies

### LINKED ISSUE #602: Please run pip check on the "conda" based tests
I noticed in https://github.com/conda-forge/pandas-gbq-feedstock/pull/47 that we don't have the necessary version of google-cloud-bigquery-storage on conda-forge, but our conda-based tests were still passing. We should catch when our dependencies need to be updated in conda-forge here.

kiraksi:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-bigquery-pandas/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes #602 🦕


chalmerlowe (changes requested):
Please add requested comment, thanks!

noxfile.py
```diff
@@ -614,6 +615,10 @@ def conda_test(session):
     # for troubleshooting purposes.
     session.run("mamba", "list")
 
+    subprocess.run(
```
chalmerlowe, line 618:
```suggestion

    # Using subprocess.run() instead of session.run() because
    # session.run() does not correctly handle the pip check command.
    subprocess.run(
```

chalmerlowe (approved):
LGTM

PR #735, opened 2024-02-23
feat: move bqstorage to extras and add debug capability

### LINKED ISSUE #718: feat: make google-cloud-bigquery-storage an optional dependency
when installing pandas-gbq, google-cloud-bigquery-storage must be installed. however this package is not a must and causes errors for clients such as permissions problems. for backwards compatibility the best possible solution is to add this package to the extras package so to be able to opt out of this package.

chalmerlowe:
Moves `google-cloud-bigquery-storage` to the `extras` section to help decrease the installation footprint.

Also adds `python -m pip freeze` to one of the nox sessions to help with debugging.

Lastly, removes two tests that have outlasted their usefulness.

Fixes #718 🦕


kiraksi (approved):
LGTM

PR #737, opened 2024-02-26
fix: Remove python 3.7 due to end of life (EOL)

chalmerlowe:
Removes support for Python 3.7 due to end of life (EOL)


chalmerlowe:
@parthea 

I seem to recall that there are some other steps in involved (was it something to do with pushing a CL to our internal repo?).

Any help you can offer is appreciated.

parthea (commented):
Also need to remove python 3.7 in sync-repo-settings.yaml for both unit tests and samples

https://github.com/googleapis/python-bigquery-pandas/blob/366cb55a5f2dde4f92994de9ba6c59eb3e1d7c9f/.github/sync-repo-settings.yaml#L14

parthea (commented):
Also update the versions here : 

https://github.com/googleapis/python-bigquery-pandas/blob/366cb55a5f2dde4f92994de9ba6c59eb3e1d7c9f/samples/snippets/noxfile.py#L92

parthea:
I just noticed that https://github.com/googleapis/python-bigquery-pandas/blob/main/testing/constraints-3.7.txt could be removed as well (and the contents moved to https://github.com/googleapis/python-bigquery-pandas/blob/main/testing/constraints-3.8.txt). 

More information about the constraints files
```
Pip uses constraints files to add restrictions when resolving dependencies. In our case, we use constraints files to ensure that the lower bound that we advertise is actually checked in a test session. Some libraries also use constraints files to ensure that a wider range of dependency versions are covered across test sessions. Usually, lower bounds are checked in the lowest supported Python version. Constraints files are kept in sync manually with setup.py. If you forget to update the constraints file after updating a dependency's minimum version in setup.py, the test session fails with a constraints error.
```

chalmerlowe:
@parthea 

Neither of the 3.7 tests is expected or needed to pass.
All other tests have passed.
Can you review and approve?

testing/constraints-3.8.txt
```diff
@@ -1 +1,20 @@
...
+google-auth==2.13.0
+google-auth-oauthlib==0.7.0
+google-cloud-bigquery==3.3.5
+google-cloud-bigquery-storage==2.16.2
 numpy==1.17.5
+pandas==1.1.4
+pyarrow==3.0.0
+pydata-google-auth==1.5.0
+tqdm==4.23.0
+protobuf==3.19.5
```
parthea, line 19: This dependency is not listed in `setup.py`. Is it a direct dependency ?

testing/constraints-3.8.txt
```diff
@@ -1 +1,20 @@
...
+#
+# e.g., if setup.py has "foo >= 1.14.0, < 2.0.0dev",
+# Then this file should have foo==1.14.0
+db-dtypes==1.0.4
+google-api-core==2.10.2
+google-auth==2.13.0
+google-auth-oauthlib==0.7.0
+google-cloud-bigquery==3.3.5
+google-cloud-bigquery-storage==2.16.2
 numpy==1.17.5
```
parthea, line 14: `numpy==1.16.6` to match the minimum supported version in `setup.py`

testing/constraints-3.8.txt
```diff
@@ -1 +1,20 @@
...
+pandas==1.1.4
+pyarrow==3.0.0
+pydata-google-auth==1.5.0
+tqdm==4.23.0
+protobuf==3.19.5
+packaging==20.0.0
```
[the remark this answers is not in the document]
chalmerlowe: I think I am a little confused.
setup.py uses setuptools to handle namespaces.
The contraint.txt files are simply to ID the lowest acceptable version of a package. And in this case we don't specify one for setuptools (relatedly, in other constraint.txt files we leave out a number of files and choose to not specify a lower bound.)

I suggest that setuptools is required for this library, but it is not critical for us to identify a minimum version boundary.

parthea (approved):

PR #768, opened 2024-05-13
fix: handle None when converting numerics to parquet

### LINKED ISSUE #719: nullable numeric column does not handled well
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/python-bigquery-pandas/issues
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - `pandas-gbq` version: 0.19.2

#### Steps to reproduce

 Try `to_gbq` with nullable Decimal rows. It does not handle None or pd.NA since it always tries to cast value into Decimal whatever it is None or not: https://github.com/googleapis/python-bigquery-pandas/blob/main/pandas_gbq/load.py#L61-L109

#### Code example

```python
df = pd.DataFrame(results)
df.to_gbq(
    'table,
    'namespace',
    if_exists='append',
    credentials=credentials,
)
```

Linchin:
`decimal.Decimal` reports an error when the input is `None`. This PR adds a lambda func to handle this case, as well as a unit test.

Fixes #719 🦕


pandas_gbq/load.py
```diff
@@ -99,7 +99,15 @@ def cast_dataframe_for_parquet(
                 errors="ignore",
             )
         elif column_type in {"NUMERIC", "DECIMAL", "BIGNUMERIC", "BIGDECIMAL"}:
-            cast_column = dataframe[column_name].map(decimal.Decimal)
+            # decimal.Decimal does not support `None` input, add support here.
+            # https://github.com/googleapis/python-bigquery-pandas/issues/719
+            def convert(x):
+                if x is None:
```
chalmerlowe, line 105: QUESTION:

The [original issue](https://github.com/googleapis/python-bigquery-pandas/issues/719) mentions that `None` and `pd.NA` are not handled appropriately.

This code seems to only address `None`.
Is the intent to check for both?

IF yes... this may be a possible check that is more general and as other edge cases arise could enable edits to the code.
```suggestion
                if x in {None, pd.NA}:
```

Linchin: Indeed, thank you for catching this! I'll update the code.

tests/unit/test_load.py
```diff
@@ -370,3 +370,24 @@ def test_cast_dataframe_for_parquet_w_null_fields():
...
+        },
+    )
+
+    schema = {"fields": [{"name": "A", "type": "BIGNUMERIC"}]}
+    result = load.cast_dataframe_for_parquet(dataframe, schema)
+
+    # pandas.testing.assert_frame_equal() doesn't distinguish Decimal('nan')
+    # vs. None, verify Decimal("nan") directly.
+    # https://github.com/pandas-dev/pandas/issues/18463
+    assert result["A"][1].is_nan()
```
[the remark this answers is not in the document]
Linchin: Thanks, I added the unit test for this case :)

chalmerlowe (approved):
LGTM

Thanks for adding the extra test case.

PR #769, opened 2024-05-16
feat: `read_gbq` suggests using BigQuery DataFrames with large results

tswast:

Fixes internal issue 340937138
🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -381,7 +381,6 @@ def cover(session):
         "report",
         "--show-missing",
         "--include=tests/unit/*",
-        "--include=tests/system/small/*",
         "--fail-under=100",
     )
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -375,6 +375,16 @@ def cover(session):
     session.install("coverage", "pytest-cov")
     session.run("coverage", "report", "--show-missing", "--fail-under=96")
 
+    # Make sure there is no dead code in our test directories.
+    session.run(
+        "coverage",
+        "report",
+        "--show-missing",
+        "--include=tests/unit/*",
+        "--include=tests/system/small/*",
+        "--fail-under=100",
+    )
+
     session.run("coverage", "erase")
 
 
--- a/pandas_gbq/constants.py
+++ b/pandas_gbq/constants.py
@@ -2,6 +2,10 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
+# BigQuery uses powers of 2 in calculating data sizes. See:
+# https://cloud.google.com/bigquery/pricing#data The documentation uses
+# GiB rather than GB to disambiguate from the alternative base 10 units.
+# https://en.wikipedia.org/wiki/Byte#Multiple-byte_units
 BYTES_IN_KIB = 1024
 BYTES_IN_MIB = 1024 * BYTES_IN_KIB
 BYTES_IN_GIB = 1024 * BYTES_IN_MIB
--- a/pandas_gbq/constants.py
+++ b/pandas_gbq/constants.py
@@ -1,0 +1,8 @@
+# Copyright (c) 2024 pandas-gbq Authors All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+BYTES_IN_KIB = 1024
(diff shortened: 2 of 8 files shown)


tswast:
`tests/system/test_read_gbq_with_bqstorage.py::test_large_results` looks like a real failure. Investigating now.

pandas_gbq/gbq.py
```diff
@@ -1050,12 +1088,7 @@ def to_gbq(
         )
 
     if api_method == "default":
-        # Avoid using parquet if pandas doesn't support lossless conversions to
```
tswast, line 1053: per coverage report, this was dead code. Our minimum pandas version is beyond the one where this feature wasn't available.

noxfile.py
```diff
@@ -375,6 +375,15 @@ def cover(session):
...
 
+    # Make sure there is no dead code in our test directories.
+    session.run(
+        "coverage",
+        "report",
+        "--show-missing",
+        "--include=tests/unit/*",
+        "--fail-under=100",
+    )
+
```
leahecole, line 386: Does this coverage addition make the previous coverage run obsolete?

pandas_gbq/constants.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2024 pandas-gbq Authors All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
```
leahecole, line 3: just double checking that this is normal given that we use a difference license header for other Google stuff

pandas_gbq/constants.py
```diff
@@ -0,0 +1,8 @@
...
+
+BYTES_IN_KIB = 1024
+BYTES_IN_MIB = 1024 * BYTES_IN_KIB
+BYTES_IN_GIB = 1024 * BYTES_IN_MIB
+BYTES_TO_RECOMMEND_BIGFRAMES = BYTES_IN_GIB
```
leahecole, line 8: Are these names governed by pandas? If not, can we use KB, MB, GB instead? Totally a nit and not a big deal if this is a convention for pandas.

pandas_gbq/exceptions.py
```diff
@@ -28,6 +28,10 @@ class InvalidPrivateKeyFormat(ValueError):
     """
 
 
+class LargeResultsWarning(UserWarning):
+    """Raise when results are beyond that recommended for pandas DataFrame."""
+
+
```
leahecole, line 34: nit - is there a threshold when this should be used? or is it up to our discretion? If specific guidance/threshold please add!

pandas_gbq/gbq.py
leahecole: thank you this is a helpful comment!

noxfile.py
tswast: No, this one is only for our test code. We still aren't quite at 100% in the actual library code. This makes sure we actually run all of the tests. It's easy to forget a `test_` prefix or have some try/except block that never actually gets run.

pandas_gbq/constants.py
tswast: Not normal, but expected. pandas-gbq was originally split off from pandas itself, so we can't use the Google header. It's not 100% Google's copyright since not all contributions were under the CLA.

pandas_gbq/constants.py
tswast: Not by pandas, but BigQuery. See: https://cloud.google.com/bigquery/pricing#storage and https://cloud.google.com/bigquery/pricing#data

> The size of your stored data and the size of the data processed by your queries is calculated in [gibibytes (GiB)](https://en.wikipedia.org/wiki/Gibibyte), where 1 GiB is 230 bytes (1,024 MiB). Similarly, 1 tebibyte (TiB) is 240 bytes (1,024 GiB).

pandas_gbq/exceptions.py
tswast: Up to our discretion. We estimate this will be reached in maybe 5% of pandas-gbq usage, so hopefully won't be too annoying and 1 GB is quite large for a pandas DataFrame, but some posts say 100 GB is the hard limit for a pandas DataFrame (I haven't found an authoritative source for that, though).

leahecole (approved):

PR #776, opened 2024-05-28
fix: clean up code that displays versions used in testing for debug purposes

chalmerlowe:
Cleaned up several items in the noxfile to faciliate debugging.
* Due to additions of `python -m pip freeze` commands, removed several package-specific version display expressions since they are now redundant.
* Removed a duplicative line with `python -m pip freeze`

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -302,7 +302,7 @@ def prerelease(session):
         "--pre",
         "--upgrade",
         "google-api-core",
-        "google-cloud-bigquery",
+        "google-cloud-bigquery==3.20.1",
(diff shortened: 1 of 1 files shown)


Linchin:
Thank you, why are we pinning google-cloud-bigquery for the pre-release test? Also the change doesn't seem to match the description.

chalmerlowe:
> BLOCKER: Thank you, why are we pinning google-cloud-bigquery for the pre-release test? Also the change doesn't seem to match the description.

Yeah. this is broken.
Gonna close this PR entirely.

PR #783, opened 2024-06-10
test: upgrading the ci/cd pipeline to reduce turnover time.

### LINKED ISSUE #778: chore: revise the CI/CD pipeline functionality in configs and noxfile
Several changes were made in the `python-bigquery` library that I would like to see replicated here:

- [x] Ensure that each major session in the `noxfile.py` has a `session.run("python", "-m", "pip", "freeze")` expression.
- [x] If necessary, remove any extraneous expressions that produce version numbers (ie. `print(pandas.__version__)`) since they will be handled by the more robust `pip freeze` expression.
- [ ] Remove `os.environ.get("RUN_SYSTEM_TESTS")` style expressions based on checking variable values in deference to using fine grained config files. See this `python-bigquery` [PR # 1915](https://github.com/googleapis/python-bigquery/pull/1915/files) for an example. (NOTE: there will also need to be some changes done in google3 internally)
- [ ] Revise the config files to separate out specific `nox.sessions` for parallel processing. Candidates include presubmit/linting-typing.cfg, presubmit/snippets, continuous/python-3.9 to 3.11) There may be several others. See here for an [example](https://github.com/googleapis/python-bigquery/blob/main/.kokoro/presubmit/linting-typing.cfg).  
- [x] Add a `@_calculate_duration` decorator to each `nox.session` to allow us to monitor how long each session actually takes. See this [PR #1917](https://github.com/googleapis/python-bigquery/pull/1917/files) for an example.

chalmerlowe:
Fixes #778  🦕

### DIFF
--- a/.kokoro/presubmit/prerelease-deps.cfg
+++ b/.kokoro/presubmit/prerelease-deps.cfg
@@ -1,0 +1,7 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Only run this nox session.
+env_vars: {
+    key: "NOX_SESSION"
+    value: "prerelease_deps"
+}
--- a/.kokoro/presubmit/prerelease-deps.cfg
+++ b/.kokoro/presubmit/prerelease-deps.cfg
@@ -1,7 +1,0 @@
-# Format: //devtools/kokoro/config/proto/build.proto
-
-# Only run this nox session.
-env_vars: {
-    key: "NOX_SESSION"
-    value: "prerelease_deps"
-}
--- a/noxfile.py
+++ b/noxfile.py
@@ -98,6 +98,7 @@ def _calculate_duration(func):
 
     return wrapper
 
+
 # 'docfx' is excluded since it only needs to run in 'docs-presubmit'
 nox.options.sessions = [
     "unit",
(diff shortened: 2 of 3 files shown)


Linchin:
QUESTION / BLOCKER:

Thanks for making ci/cd more efficient! Just to double check, are we getting rid of pre-release tests altogether? Also owlbot added something. I'm not sure if this came from some new internal change, or we need to change owlbot.py.

chalmerlowe:
> QUESTION / BLOCKER:
> 
> Thanks for making ci/cd more efficient! Just to double check, are we getting rid of pre-release tests altogether?

There were two versions of prerelease testing.
* `prerelease` (this one is intended to remain)
* `prerelease-deps` (this was basically a duplicate AND it never runs (because it is not part of the google3 configuration setup) so I attempted to delete it.

> Also owlbot added something. I'm not sure if this came from some new internal change, or we need to change owlbot.py.

I looked at the owlbot commit. 
* It added back in the prerelease-deps file that I deleted. I will go fix that. 
* AND it cleaned up a line spacing issue.

Linchin:
LGTM. Thank you!

Linchin (approved):

PR #817, opened 2024-10-01
chore: adds Python 3.7/3.8 EOL pending deprecation warning

chalmerlowe:
Adds a `PendingDeprecationWarning` and additional code logic to output that warning if the user attempts to install the repo using Python 3.7 or 3.8.

The expected outcome is that the system will display a PendingDeprecationWarning if it detects Python runtimes 3.7 or 3.8, and will display nothing if using more recent Python runtimes 3.9 to 3.x.

### DIFF
--- a/pandas_gbq/__init__.py
+++ b/pandas_gbq/__init__.py
@@ -6,6 +6,19 @@ from pandas_gbq import version as pandas_gbq_version
 
 from .gbq import Context, context, read_gbq, to_gbq  # noqa
 
+from . import _versions_helpers
+
+sys_major, sys_minor, sys_micro = _versions_helpers.extract_runtime_version()
+if sys_major == 3 and sys_minor in (7, 8):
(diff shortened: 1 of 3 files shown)


pandas_gbq/__init__.py
```diff
@@ -2,10 +2,24 @@
...
 # license that can be found in the LICENSE file.
 
+import warnings
+
 from pandas_gbq import version as pandas_gbq_version
 
+from . import _versions_helpers
 from .gbq import Context, context, read_gbq, to_gbq  # noqa
 
+sys_major, sys_minor, sys_micro = _versions_helpers.extract_runtime_version()
```
Linchin, line 12:
```suggestion
sys_major, sys_minor, _ = _versions_helpers.extract_runtime_version()
```
a nit suggestion

Linchin (approved):

PR #838, opened 2024-12-12
fix: `to_gbq` uses `default_type` for ambiguous array types and struct field types

### LINKED ISSUE #836: AttributeError: 'NoneType' object has no attribute 'to_api_repr'
In the new pandas_gqb Version 0.24.0 there seems to be a bug with writing to BigQuery tables from dataframes where there are nested columns which only include NULL values. However the Schema type should be clear since we provide a Schema for type information.

With pandas_gbq Version=0.24.0 - we get a "AttributeError: 'NoneType' object has no attribute 'to_api_repr' " Error - for the previous version 0.23.0 we do not get this error and the script works.

I assume the reason is that in the library we try to get the schema for this dataframe in "gbq.py" eventhough the schema would be provided.  The function "generate_bq_schema" fails eventhough the execution would not be necessary in this case:
```
  default_schema = _generate_bq_schema(dataframe) --> Throws the error since it can not infer the schema
  # If table_schema isn't provided, we'll create one for you
  if not table_schema:
      table_schema = default_schema
  # It table_schema is provided, we'll update the default_schema to the provided table_schema
  else:
      table_schema = pandas_gbq.schema.update_schema(
          default_schema, dict(fields=table_schema)
      )
```

#### Environment details

  - Python version: 3.12.7
  - pip version: 24.2
  - `pandas-gbq` version: 0.24.0
  - pandas version: 2.2.3
  - numpy version: 1.26.4

#### Steps to reproduce

  1. Execute script below with the above versions -> Fails.
  2. Execute script below with above versions - but adjust pandas-gbq==0.23.0 -> Works

#### Code example

```python
import pandas_gbq
import pandas as pd
import numpy as np

DESTINATION_TABLE_ID = 'INSERT_YOUR_TABLE_HERE'

schema = [
 {'name': 'Id', 'type': 'INTEGER', 'mode': 'NULLABLE'},
 {'name': 'Positions',
  'type': 'RECORD',
  'mode': 'REPEATED',
  'fields': [
   {'name': 'PositionState',
    'type': 'STRING',
    'mode': 'NULLABLE'}
  ]
}
]

works_df = pd.DataFrame([{
        'Id': 123,
        'Positions': None
}])

error_df = pd.DataFrame([{
        'Id': 123,
        'Positions': np.array([{
            'PositionState': None
        }])
}])

# Works with warning
# pandas_gbq.to_gbq(works_df, destination_table=DESTINATION_TABLE_ID, table_schema=schema, if_exists='replace')

# Throws error
pandas_gbq.to_gbq(error_df, destination_table=DESTINATION_TABLE_ID, table_schema=schema, if_exists='replace')
```

#### Stack trace
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[20], line 2
      1 import pandas_gbq
----> 2 pandas_gbq.to_gbq(dummy_df, destination_table='DESTINATION_TABLE_ID', table_schema=schema)

File ~/anaconda3/envs/py-312/lib/python3.12/site-packages/pandas_gbq/gbq.py:1163, in to_gbq(dataframe, destination_table, project_id, chunksize, reauth, if_exists, auth_local_webserver, table_schema, location, progress_bar, credentials, api_method, verbose, private_key, auth_redirect_uri, client_id, client_secret, user_agent, rfc9110_delimiter)
   1160 dataset_id = destination_table_ref.dataset_id
   1161 table_id = destination_table_ref.table_id
-> 1163 default_schema = _generate_bq_schema(dataframe)
   1164 # If table_schema isn't provided, we'll create one for you
   1165 if not table_schema:

File ~/anaconda3/envs/py-312/lib/python3.12/site-packages/pandas_gbq/gbq.py:1249, in _generate_bq_schema(df, default_type)
   1246 fields_json = []
   1248 for field in fields:
-> 1249     fields_json.append(field.to_api_repr())
   1251 return {"fields": fields_json}

File ~/anaconda3/envs/py-312/lib/python3.12/site-packages/google/cloud/bigquery/schema.py:353, in SchemaField.to_api_repr(self)
    350 # If this is a RECORD type, then sub-fields are also included,
    351 # add this to the serialized representation.
    352 if self.field_type.upper() in _STRUCT_TYPES:
--> 353     answer["fields"] = [f.to_api_repr() for f in self.fields]
    355 # Done; return the serialized dictionary.
    356 return answer

AttributeError: 'NoneType' object has no attribute 'to_api_repr'
```

tswast:
TODO:

- [n/a] Appropriate docs were updated (if necessary)
- [n/a] ~Include fix to not detect schema if one is provided (or merge with the provided one if need be).~ N/A. There is merge logic so that users can provide a subset of the schema. I don't think it makes sense to modify that logic in this PR.

Fixes #836 🦕


pandas_gbq/schema/pandas_to_bigquery.py
```diff
@@ -164,7 +166,13 @@ def dtype_to_bigquery_field(name, dtype) -> Optional[schema.SchemaField]:
     return None
 
 
-def value_to_bigquery_field(name, value) -> Optional[schema.SchemaField]:
+def value_to_bigquery_field(
+    name, value, default_type=None
+) -> Optional[schema.SchemaField]:
+    # There are no non-null values, so assume the default type.
```
chalmerlowe, line 172:
```suggestion
    # Set the SchemaField datatype to the given default_type if the value
    # being assessed is None.
```

tswast: Thanks! This does sound much clearer.

pandas_gbq/schema/pandas_to_bigquery.py
```diff
@@ -188,29 +197,33 @@ def value_to_bigquery_field(name, value) -> Optional[schema.SchemaField]:
     return None
 
 
-def values_to_bigquery_field(name, values) -> Optional[schema.SchemaField]:
+def values_to_bigquery_field(
+    name, values, default_type="STRING"
+) -> Optional[schema.SchemaField]:
```
chalmerlowe, line 202: **#PREFERENCE**: See note above about docstrings.

```suggestion
def values_to_bigquery_field(
    name: str, values: Any, default_type: str = "STRING"
) -> Optional[schema.SchemaField]:
    """Infers the BigQuery schema field type from a list of values.

    This function iterates through the given values to determine the
    corresponding schema field type.

    Args:
        name:
            The name of the field.
        values:
            An iterable of values to infer the type from. If all the values
            are None or the iterable is empty, the function returns None.
        default_type:
            The default field type to use if a specific type cannot be
            determined from the values. Defaults to "STRING".

    Returns:
        The schema field, or None if a type cannot be inferred.
    """
```

pandas_gbq/schema/pandas_to_bigquery.py
```diff
@@ -188,29 +197,33 @@ def value_to_bigquery_field(name, value) -> Optional[schema.SchemaField]:
...
     value = pandas_gbq.core.pandas.first_valid(values)
 
-    # All NULL, type not determinable.
+    # All values came back as NULL, thus type not determinable by this method.
```
chalmerlowe, line 205: I know we are pushing to merge to enable further work, so the following is merely a #PREFERENCE or #SUGGESTION, not a blocker.

* The code here works and tests pass so I don't feel like forcing us to fix something that isn't broken, but I wonder if we could clean up the code slightly (better readability, maintainability, efficiency). 
  * we use `isinstance(v, str)` in two functions despite the fact that the outer function calls the inner function, which does a string check and then the outer function immediately does a second `isinstance()` check.
  * we use a `field is not None` double negative style check when I suspect we can safely do an `if field` check.

Will leave it up to you to decide whether any of these changes are reasonable in this PR.

```
    value = pandas_gbq.core.pandas.first_valid(values)

    # All values came back as NULL, thus type not determinable by this method.
    # Return None so we can try other methods.
    if value is None:
        return None

    field = value_to_bigquery_field(name, value, default_type=default_type)
    if field:
        return field

    # Check plain ARRAY values. Exclude mapping types to allow pyarrow to determine
    # STRUCT which can examine more values to determine all the keys
    if isinstance(value, collections.abc.Iterable) and not isinstance(
        value, collections.abc.Mapping
    ):
        # This collection may contain only None elements or is empty, so get the
        # first non-None value we can find.
        valid_item = pandas_gbq.core.pandas.first_array_valid(values)
        field = value_to_bigquery_field(name, valid_item, default_type=default_type)
        if field:
            return schema.SchemaField(name, field.field_type, mode="REPEATED")

    return None
```

pandas_gbq/schema/pandas_to_bigquery.py
```diff
@@ -164,7 +166,14 @@ def dtype_to_bigquery_field(name, dtype) -> Optional[schema.SchemaField]:
     return None
 
 
-def value_to_bigquery_field(name, value) -> Optional[schema.SchemaField]:
+def value_to_bigquery_field(
+    name, value, default_type=None
+) -> Optional[schema.SchemaField]:
+    # Set the SchemaField datatype to the given default_type if the value
```
chalmerlowe, line 172: I know we are pushing to merge to enable further work, so the following is merely a #PREFERENCE or #SUGGESTION, not a blocker.

The code here works and tests pass so I don't feel like forcing us to fix something that isn't broken, but I wonder if we could clean up the code slightly (better readability, maintainability, efficiency).

* a type mapping removes some duplicate `isinstance()` calls and makes it easy to extend the code if we need to add checks against additional datatypes
* as seen in a comment that follows, we use `isinstance(v, str)` in two functions despite the fact that the outer function calls the inner function, which does a string check and then the outer function immediately does a second `isinstance()` check.

Will leave it up to you to decide whether any of these changes are reasonable in this PR.

```    if value is None:
        return schema.SchemaField(name, default_type)

    type_mapping = {
        str: "STRING",
        _BaseGeometry: "GEOGRAPHY",
    }

    for type_, field_type in type_mapping.items():
        if isinstance(value, type_):
            return schema.SchemaField(name, field_type)

    # For timezone-naive datetimes, the later pyarrow conversion to try and
    # learn the type add a timezone to such datetimes, causing them to be
    # recognized as TIMESTAMP type. We thus additionally check the actual data
    # to see if we need to overrule that and choose DATETIME instead.
    #
    # See: https://github.com/googleapis/python-bigquery/issues/985
    # and https://github.com/googleapis/python-bigquery/pull/1061
    # and https://github.com/googleapis/python-bigquery-pandas/issues/450
    if isinstance(value, datetime.datetime):
        return schema.SchemaField(
            name, "TIMESTAMP" if value.tzinfo else "DATETIME"
        )

    return None
```

pandas_gbq/schema/pyarrow_to_bigquery.py
```diff
@@ -37,7 +37,15 @@
 }
 
 
-def arrow_type_to_bigquery_field(name, type_) -> Optional[schema.SchemaField]:
+def arrow_type_to_bigquery_field(
```
chalmerlowe, line 40: please provide a docstring.

chalmerlowe (approved):
- [x] Please accept and/or add docstrings. 
- [x] Please consider the other suggestions for inclusion, but not blockers.

Approving this PR based on addition of docstrings.

pandas_gbq/schema/pandas_to_bigquery.py
tswast: Good point. Removed the redundant string check.

pandas_gbq/schema/pyarrow_to_bigquery.py
tswast: Added

PR #880, opened 2025-02-14
chore: Add docs confit path to .readthedocs.yml

Linchin:
Fixes test failures such as https://readthedocs.org/projects/pandas-gbq/builds/27154263/


Linchin:
use #881 instead

PR #891, opened 2025-03-05
build: use python 3.10 for testing

parthea:
This should resolve the following issue in the `prerelease` nox session

```
ERROR: Package 'google-cloud-bigquery' requires a different Python: 3.8.20 not in '>=3.9'
nox > Session prerelease failed.
```


Linchin (approved):
LGTM, thanks a lot for fixing the test

Linchin:
@parthea  Looks like we need to update the template for owlbot instead?

PR #909, opened 2025-04-22
refactor: break down gbq.py file to several smaller ones

sycai:
The intention of this PR is to re-organize the code to make it easier to implement the dry_run feature: https://github.com/googleapis/google-cloud-python/issues/14485

The file `gbq.py` contains too many lines which makes it hard for code navigations.

Change summary:
* `GbqConnector` and its helper functions are relocated from `gbq.py` to a new file `gbq_connector.py`
* Exceptions defined in `gbq.py` are relocated to `exceptions.py` file. 
* `Context` class and `context` global instance are relocated from `gbq.py` to a new file `context.py`. 

These items are all imported back to the `gbq.py` file to avoid breaking imports of our downstream products.


pandas_gbq/gbq.py
```diff
@@ -6,32 +6,20 @@
...
-# dependencies in the pandas package, which has an optional dependency on
-# pandas-gbq.
-if typing.TYPE_CHECKING:  # pragma: NO COVER
-    import pandas
-
-import pandas_gbq.constants
-import pandas_gbq.exceptions
-from pandas_gbq.exceptions import GenericGBQException, QueryTimeout
+from pandas_gbq.contexts import context
+from pandas_gbq.exceptions import (
```
tswast, line 12: Let's be explicit about what aliases we are keeping around with `__all__` + a comment explaining why these "unused" aliases are here.

sycai: `__all__` may give us extra burden due to potential star imports. I used `# noqa` to mark all unused imports instead.

pandas_gbq/gbq_connector.py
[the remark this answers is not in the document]
sycai: As discussed offline. We won't address this issue for now.

That said, GbqConnector does seem like an over-engineered OOP class.

chalmerlowe (approved):
LGTM. Approving, pending resolution of failing tests in prerelease CI/CD checks.

PR #927, opened 2025-06-03
fix: remove pandas-gbq client ID for authentication

tswast:
pandas-gbq's client ID has been failing for some time due to `Error 400: redirect_uri_mismatch`. Since it's only used for a fallback if no application default credentials are available, it's probably OK to use pydata-google-auth's client ID instead.

Fixes #<issue_number_goes_here> 🦕


mrfaizal (approved):
lgtm assuming we're ok removing a fallback option. since it wasn't working anyways, this is no worse than before.

shuoweil (approved):

PR #949, opened 2025-09-22
feat: Add partitioning and clustering to the to_gbq function

antoineeripret:
Full context: https://github.com/googleapis/google-cloud-python/issues/14488

The objective of this change is to ensure that we can leverage partitioning and clustering when using the `pands_gbq` library.  New arguments added to the function (and documented in the code). 

```python 
clustering_columns
time_partitioning_column
time_partitioning_type
time_partitioning_expiration_ms
range_partitioning_column
range_partitioning_range
```

This allows us to execute the following code (for example): 

```python 

pandas_gbq.to_gbq(
    df, 
    "xxxxx.yyyy", 
    project_id="zzzzz",
    if_exists="replace", 
    credentials=credentials, 
    time_partitioning_column="date", 
    clustering_columns=['country', 'page']
)
```

Which would end up in the following configuration being applied in BigQuery: 

image

### DIFF
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -10,6 +10,7 @@ import typing
 import warnings
 
 import pandas
+
 from pandas_gbq.contexts import Context  # noqa - backward compatible export
 from pandas_gbq.contexts import context
 from pandas_gbq.exceptions import (  # noqa - backward compatible export
@@ -797,11 +798,11 @@ class _Table(GbqConnector):
         """
         from google.cloud.bigquery import (
             DatasetReference,
+            PartitionRange,
+            RangePartitioning,
             Table,
             TableReference,
             TimePartitioning,
-            RangePartitioning,
-            PartitionRange,
         )
 
         if self.exists(table_id):
--- a/pandas_gbq/gbq.py
+++ b/pandas_gbq/gbq.py
@@ -416,9 +416,7 @@ def to_gbq(
     destination_table : str
         Name of table to be written, in the form ``dataset.tablename`` or
         ``project.dataset.tablename``.
-    clustering_columns : typing.Union[
-        pandas.core.indexes.base.Index, typing.Iterable[typing.Hashable]
-    ] = (),
+    clustering_columns : pandas.Index | Iterable[Hashable], optional
         Specifies the columns for clustering in the BigQuery table.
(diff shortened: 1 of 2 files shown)


tswast (commented):
We try to keep unit test coverage high. Would you mind adding a unit test or two or three similar to

https://github.com/googleapis/python-bigquery-pandas/blob/48a91df1c2b986ec97edd07bb513c3e965ab94b6/tests/unit/test_gbq.py#L306-L323

that confirms that these options are passed through to the `create_table` call as expected?

pandas_gbq/gbq.py
```diff
@@ -762,6 +824,26 @@ def create(self, table_id, schema):
         table = Table(table_ref)
         table.schema = pandas_gbq.schema.to_google_cloud_bigquery(schema)
 
+        if clustering_columns:
+            table.clustering_fields = list(clustering_columns)
+
+        if time_partitioning_column:
```
tswast: Technically someone could enable time partitioning without a column to do [ingestion time partitioning](https://docs.cloud.google.com/bigquery/docs/partitioned-tables#ingestion_time). This is less flexible than using a column, though, so I'm not sure we need to worry about it.

tswast (dismissed):
Thank you!

pandas_gbq/gbq.py
```diff
@@ -406,6 +416,20 @@ def to_gbq(
     destination_table : str
         Name of table to be written, in the form ``dataset.tablename`` or
         ``project.dataset.tablename``.
+    clustering_columns : typing.Union[
+        pandas.core.indexes.base.Index, typing.Iterable[typing.Hashable]
+    ] = (),
```
tswast: Getting a docs failure. I think it might be this line.

```
sphinx.errors.SphinxWarning: /tmpfs/src/github/python-bigquery-pandas/pandas_gbq/gbq.py:docstring of pandas_gbq.gbq.to_gbq:19:Field list ends without a blank line; unexpected unindent.
```

Let's try this.

```suggestion
    clustering_columns : pandas.Index | Iterable[Hashable]
```

antoineeripret:
@tswast : added  your suggestion in the last commit, thanks !

tswast (approved):
Thank you!

shuoweil (approved):

PR #979, opened 2025-11-06
feat: add dry run to the read_gbq function

antoineeripret:
This change allows the user to run a [dry run query](https://docs.cloud.google.com/bigquery/docs/samples/bigquery-query-dry-run?hl=en) using the `read_gbq` function. Instead of returning a `pd.DataFrame`, the behavior is changed and the amount of data processed (in GB) is returned.


antoineeripret:
Hi @shuoweil, the last commit should fix it. Got the following on my local env: 

```bash
python -m black --check docs pandas_gbq tests noxfile.py setup.py
All done! ✨ 🍰 ✨
45 files would be left unchanged.
```

pandas_gbq/gbq_connector.py
```diff
@@ -260,6 +268,20 @@ def run_query(self, query, max_results=None, progress_bar_type=None, **kwargs):
...
+            # RowIterator has a job attribute that references the QueryJob
+            query_job = (
+                rows_iter.job if hasattr(rows_iter, "job") and rows_iter.job else None
+            )
+            if query_job is None:
+                # Fallback: if query_and_wait_via_client_library doesn't set job,
+                # we need to get it from the query result
+                # For query_and_wait_via_client_library, the RowIterator should have job set
+                raise ValueError("Cannot access QueryJob from RowIterator for dry_run")
+            return query_job.total_bytes_processed / 1024**3
```
sycai: Could we simply return `query_job.total_bytes_processed` without further processing? 

Reasons:
* The `total_bytes_processed` has integer type, which is more precise than a float type
* For small tables (ones with 1-10 MB sizes), converting the size to GB makes the result less readable
* It aligns more with the behavior of BigQuery Python client to return size in bytes.

Generally speaking, we want the caller of this function to perform unit conversions.

tests/unit/test_gbq.py
```diff
@@ -937,3 +939,20 @@ def test_run_query_with_dml_query(mock_bigquery_client, mock_query_job):
...
+        hasattr(mock_bigquery_client, "query_and_wait")
+        and mock_bigquery_client.query_and_wait.called
+    ):
+        _, kwargs = mock_bigquery_client.query_and_wait.call_args
+        job_config = kwargs["job_config"]
+    else:
+        _, kwargs = mock_bigquery_client.query.call_args
+        job_config = kwargs["job_config"]
+    assert job_config.dry_run is True
+    assert cost == 12345 / 1024**3
```
sycai: I guess we will have to fix this assertion too.

sycai (commented):
Thank you! I think we should be good to go once the doc and tests are updated.

antoineeripret:
@sycai : updated :)

sycai (dismissed):

sycai (dismissed):

shuoweil (approved):

shuoweil (changes requested):
@antoineeripret Please check the failed testcase tests/system/test_gbq.py::TestReadGBQIntegration::test_read_gbq_with_dry_run

sycai (approved):

shuoweil (approved):

